PR #3, opened 2020-09-30
Add std/alloc features

l1h3r:
This is for @rootmos to show how heap-allocation could be user-customizable. `std` and `alloc` features are common in Rust

- `std` would be used to enable features that include functionality from the stdlib; this is useful for integration with `std::error::Error` (and currently a requirement of the `thiserror` crate). 

- `alloc` would be used for features requiring heap allocations (eg. RSA). Users would implement a global allocator with the [`GlobalAlloc` trait](https://doc.rust-lang.org/core/alloc/trait.GlobalAlloc.html)

---

heap-allocating structures are now imported like this:

```rust
#[cfg(feature = "alloc")]
use alloc::vec::Vec;
```
---

Here's an example allocator that is often used for WebAssembly: [`wee_alloc`](https://github.com/rustwasm/wee_alloc)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,9 @@ default = ["std"]
 std = ["alloc"]
 alloc = []
 
-xchacha20poly1305 = [ "chacha20poly1305" ]
+xchacha20poly1305 = [
+  "chacha20poly1305/xchacha20poly1305",
+]
 
 [dependencies]
-chacha20poly1305 = { version = "0.6.0", optional = true }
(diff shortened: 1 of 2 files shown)


l1h3r:
> This is a great place to collect our thoughts on allocation

@rootmos do we still need to keep this open?

nothingismagick:
@rootmos - can we close it?

nothingismagick:
@i1h3r - would you mind turning this into an issue? going to close it for now because more discussion is needed.

PR #5, opened 2020-10-05
libjose signatures

l1h3r:
### TODO


nothingismagick:
Do you still want to add these to crypto.rs?

l1h3r:
> Do you still want to add these to crypto.rs?

Yes, but we can close this PR and I'll open smaller individual PRs for what we need in identity.rs

PR #10, opened 2020-11-12
feat(hash) add blake2b hashing function

elenaf9:
Implement blake2b hashing function


src/blake2b.rs
```diff
@@ -0,0 +1,42 @@
...
+
+pub fn hash(data: &[u8], output_buffer: &mut [u8]) {
+    let hash = Params::new().hash_length(output_buffer.len()).hash(data);
+    let hash_bytes = hash.as_bytes();
+    output_buffer[..hash_bytes.len()].clone_from_slice(&hash_bytes[..]);
+}
+
+#[test]
+fn test_hash_distinctness() {
+    let data_1 = b"15ed49f3b886ec5756459fdf43903c2f876ca1fb1144bb96b05e8fb116fcf7fc97fc53f1f30e350f7b304bf919b228083a8ffe1f9a0dd2162bdcb3c7a3846cc0acf04d88a5f87b32";
```
rootmos: I added a `crate::test_utils::fresh::bytestring()` that'll give you an arbitrary length bytestring that we can use here.

Wollac (commented):
Since we are essentially letting an external library doing all the work, the least thing we have to do is proper testing. The official BLAKE2 testvectors are located here: https://github.com/BLAKE2/BLAKE2/tree/master/testvectors
`blake2b_simd` happens to test exactly against these vectors, but we should still do it ourselves in case we change that implementation and to not have to trust in external tests. 
Ideally, the BLAKE2 repo should be included as a sub-module and the test vectors loaded from one of the provided files, but copying the test vectors (similar to https://github.com/oconnor663/blake2_simd/tree/master/tests).
Unfortunately, these tests are all keyed 512-bit hashes, which in some sense is a superset of the  unkeyed BLAKE2b-256 that is used in the protocol. Therefore, there should also be some BLAKE2b-256 tests, maybe against `b2sum --length 256` or similarly accepted libraries.

Wollac (approved):
Looks good to me!
With the added tests we should be on the safe side!

PR #27, opened 2020-12-10
Correct Ed25519-SHA512 nomenclature

rootmos:
https://github.com/ZcashFoundation/ed25519-zebra/blob/0e7a96a267a756e642e102a28a44dd79b9c7df69/src/signing_key.rs#L75

### DIFF
--- a/src/ed25519.rs
+++ b/src/ed25519.rs
@@ -25,8 +25,8 @@ impl SecretKey {
         self.0.into()
     }
 
-    pub fn from_le_bytes(bs: [u8; SECRET_KEY_LENGTH]) -> crate::Result<Self> {
-        Ok(SecretKey(ed25519_zebra::SigningKey::from(bs)))
+    pub fn from_le_bytes(bs: [u8; SECRET_KEY_LENGTH]) -> Self {
+        SecretKey(ed25519_zebra::SigningKey::from(bs))
     }
 
     pub fn sign(&self, msg: &[u8]) -> Signature {
@@ -78,6 +78,19 @@ mod tests {
         signature: &'static str,
     }
 
+    #[cfg(feature = "random")]
+    pub fn fresh_key() -> SecretKey {
+        SecretKey::generate().unwrap()
+    }
+
+    #[cfg(not(feature = "random"))]
+    pub fn fresh_key() -> SecretKey {
+        let mut bs = [0u8; SECRET_KEY_LENGTH];
+        use rand::{rngs::OsRng, RngCore};
+        OsRng.fill_bytes(&mut bs);
+        SecretKey::from_le_bytes(bs)
+    }
+
     #[test]
     fn test_vectors() -> crate::Result<()> {
         let tvs = [
@@ -105,7 +118,7 @@ mod tests {
         for tv in tvs.iter() {
             let mut skb = [0; SECRET_KEY_LENGTH];
             hex::decode_to_slice(tv.secret_key, &mut skb as &mut [u8]).unwrap();
-            let sk = SecretKey::from_le_bytes(skb)?;
+            let sk = SecretKey::from_le_bytes(skb);
             assert_eq!(skb, sk.to_le_bytes());
 
             let mut pkb = [0; COMPRESSED_PUBLIC_KEY_LENGTH];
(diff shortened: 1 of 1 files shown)


nothingismagick (approved):

Wollac:
What exactly is the main purpose of those changes?
- To clarify naming? In this case, I don't fully understand. By definition and specification Ed25519 is an EdDSA signature scheme always using SHA-512 (among other parameters) replacing _H_ in Ed25519 with something different, is no longer Ed25519 (but of course still EdDSA). 
- To support alternative Ed25519-like signature scheme using different _H_ (like maybe Blake2) in code? That is of course nice to have, but do we have any concrete applications for those schemes especially since they break compatibility?

rootmos:
I would challange the *always* by referring to: https://ed25519.cr.yp.to/ed25519-20110926.pdf, page 7, "Choice of curve": "Specifically, Ed25519-SHA-512 is EdDSA with the following parameters: b =256; H is SHA-512; ..."

and also in the paragraph before that: "We use the name Ed25519 for EdDSA with this particular choice of curve." not mentioning the choice of H.

Wollac:
Ed25519 is standardized in the IRTF RFC https://tools.ietf.org/html/rfc8032 and not in the academic paper. And here it is assured at multiple places that H is indeed set to SHA-512.
All libraries and implementations for Ed25519 should follow that particular RFC. And as such, SHA-512 is used as well as all the other EdDSA parameters proposed in the RFC.

nothingismagick:
I think being explicit is useful and doesn't change the nature of the implementation.

rootmos:
How about a compromise? This ought to make any context misunderstandings go away.

```
diff --git a/src/ed25519.rs b/src/ed25519.rs
index 83d73f4..386ba0f 100644
--- a/src/ed25519.rs
+++ b/src/ed25519.rs
@@ -32,6 +32,10 @@ impl AsRef<SecretKey> for SecretKey {
     }
 }
 
+pub mod RFC8032 {
+    pub use super::SHA512::*;
+}
+
 pub mod SHA512 {
     use super::*;
     use core::convert::TryFrom;
```

Wollac:
Don't get me wrong, making code more modular, more explicit, easier maintainable is always great and this PR does not change the functionality!

Libraries and implementations of Ed25519 follow the RFC 8032, especially because the academic papers do not unambiguously describe all parameters or edge-case. [IOTA](https://github.com/GalRogozinski/protocol-rfcs/blob/ed25519/text/0028-ed25519-validation/0028-ed25519-validation.md) as well as ed25519-zebra explicitly state at multiple locations that they follow RFC 8032. As such, I hesitated when reading the PR title or the `ed25519_zebra is implementing the Ed25519-SHA512 EdDSA scheme without explicitly stating that fact` comment.
However, I am mostly worried about the user of this library. As usage is almost as critical as the actual implementation (especially for cryptography), I am generally hoping to reduce friction by preferring "industry standard" names that people might be more used to. I am not sure, whether calling it SHA512 or RFC8032 could maybe lead to unnecessary confusion in this context. Could calling it maybe something like "standard/default" or something along these lines help?
The very least, however, what we should do is add more documentation somewhere addressing exactly these potential issues.

nothingismagick:
Regarding the "more documentation" - this is why I opened the discussions tab on this repo. A lot of things in Crypto are quite contentious, people (generally) don't read conversations like this after a PR is merged. But we definitely need to have more documentation. Would love to hear what @huhn511 has to say on that matter.

huhn511:
Hey :wave: 

im not that familiar with crypto.rs - but i think we should hold 99% of the documentation here in the repo and just describe and link to it from docs.iota.org. It's a very specific lib and most of IOTA users don't need it, right? 

With the "SHA512 or RFC8032" naming, i would suggest to "name" it and link to the specification - so users can follow the link and read about whats behind the name. But not sure if i understand you right.

rootmos:
Ok, so my comment in the code might be too sharp and I can remove it. Although it can be read and understood as conveying my belief that the naming conventions in the RFC in question is making an important choice implicit. If they do that in order to appease the "industry", that does not impact my opinion that being explicit is superior to being implicit in a cryptographic context. The naming using scopes in the code I suggest makes the choice of H explicit as well as guide applications wanting to be compliant with the RFC. From my last statement should also be inferred that I'm of the opinion that standards do not imply a universal truth, only when being explicit with what standards one wishes to be compliant with do they have meaning and only in that restricted context.

jakubcech:
> Hey 👋
> 
> im not that familiar with crypto.rs - but i think we should hold 99% of the documentation here in the repo and just describe and link to it from docs.iota.org. It's a very specific lib and most of IOTA users don't need it, right?
> 
> With the "SHA512 or RFC8032" naming, i would suggest to "name" it and link to the specification - so users can follow the link and read about whats behind the name. But not sure if i understand you right.

It's mainly used as a dependency for in other products that will be documented. It doesn't need it's own docs.iota.org place, at least for now.

nothingismagick:
I am closing this PR, because it is outdated and conflicts.

PR #31, opened 2020-12-14
BIP39 wordlist codec

rootmos:


nothingismagick (approved):
NICE!

rootmos:
The implementation is not particularly pretty all our test vectors and test-cases pass.

src/bip39.en.rs
```diff
@@ -0,0 +1,2055 @@
+// Copyright 2020 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+// https://github.com/trezor/python-mnemonic/tree/master/mnemonic/wordlist
```
Wollac:
```suggestion
// https://github.com/bitcoin/bips/blob/master/bip-0039/english.txt
```
Referencing the word lists under bitcoin/bips is maybe a bit cleaner than using the 3rd party Trezor link

src/bip39.jp.rs
```diff
@@ -0,0 +1,2055 @@
+// Copyright 2020 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+// https://github.com/trezor/python-mnemonic/tree/master/mnemonic/wordlist
```
Wollac:
```suggestion
// https://github.com/bitcoin/bips/blob/master/bip-0039/japanese.txt
```

src/bip39.rs
```diff
@@ -10,152 +10,363 @@ type Passphrase = str;
...
+                        y
+                    }
+                    _ => return Ok(ms.unwrap()),
+                }
+            };
+
+            match ms {
+                None => ms = Some(wordlist[idx].to_string()),
+                Some(ref mut ms) => {
+                    ms.push(' ');
```
Wollac: Unfortunately, Japanese is special as it uses a different [separator](https://github.com/bitcoin/bips/blob/master/bip-0039/bip-0039-wordlists.md#japanese). The test-vectors from https://github.com/bip32JP/bip32JP.github.io/blob/master/test_JP_BIP39.json contain this separator. However, since in the tests of this PR only the normalized string is compared, they can still pass.

I am not sure, how to best deal with this:
- Is it worth to make the word separator language depended?
- Only support English? Since localization is difficult and the seed does depend on the Language (when generated from entropy) almost all applications will only use English anyways.
- Ignore the Japanese space? (while this is technically not 100% correct)
- ...

src/bip39.en.rs
rootmos: Nice, I think I was reading an old version of the wiki where they referred to the trezor lists.

src/bip39.rs
rootmos: Ah, good point.

I kinda like the idea of extending the wordlist type with the separator, probably won't impact users since they'll most likely use the static lists that are defined in the same module.

rootmos: Can you help me figure out the Unicode code point(s) for the jp separator? The link you gave only have the utf-8 encoding (if I read it correctly.)

Wollac: Does this help: https://www.fileformat.info/info/unicode/char/3000/index.htm

rootmos: I keep the attempted fix here (don't have the time right now to figure out why the tests aren't passing): https://github.com/iotaledger/crypto.rs/commit/5035ebdb5b483b05bd601ecf86d371b089d03bf3

rootmos: I looked a bit at the bytes from the tests with that separator, and it seems like it's only the last one or two code points that differ. This feels like someone who has dealt with Japanese encodings before will just know. (I'm way too curious so I'm currently eyeballing the trezor implementation for clues :smile: )

But I suggest we move forward like this:
1. use the new Wordlist type that has the separator per language so that there's at least a clear way forward for someone that knows how to address this
2. only check the generated mnemonics when wordlist is not Japanese in the test vectors test, and mark that skipped par with a marked TODO
3. raise an issue that can track said TODO
4. merge if you don't see other issues

I'm not really happy with removing the Japanese test vectors completely at this stage as those are the ones that exercise the normalization in the decode/verify and mnemonic_to_seed functions. Unless you have another set of test vectors for some language more complex than English but less complex than Japanese up your sleeve?

rootmos: Did a quick comparison and the Japanese wordlist from [bitcoin/bips](https://raw.githubusercontent.com/bitcoin/bips/master/bip-0039/japanese.txt) and [bip32JP](https://github.com/bip32JP/bip32JP.github.io/blob/0d1ac71933458f08a56d6ab44dd0f251158cc865/js/bip39.js#L11) are identical, so that's not an issue.

Wollac: I remember that the Japanese test vectors also gave me a very hard time for https://github.com/Wollac/iota-crypto-demo/tree/master/pkg/bip39 (eventually I just cheated with the tests and only compared the normalized :see_no_evil: ).
@rootmos maybe this is also the issue you are having:
Voiced consonants in Japanese Hiragana are denoted by a Dakuten (゛) and in Unicode there seems to be the possibility to either encode this as a combination of two characters or as one. For example:
- https://www.fileformat.info/info/unicode/char/305e/index.htm **or**
- https://www.fileformat.info/info/unicode/char/305d/index.htm plus https://www.fileformat.info/info/unicode/char/3099/index.htm 

Changing the wordlist to use the Dakuten and similar in the matching format as the tests would fix this. However I don't know what the recommended / normalized Unicode way is! Maybe this was introduced by different encodings of wordlist vs tests on GitHub? Or you are implicitly normalizing the wordlist but not the tests...

rootmos: I actually tried to be very careful to only use wget, sed, grep to create the wordlist files, not even open them in any kind of editor. But I agree that something like that is probably the cause.

I've pushed the "half-way" solution now: https://github.com/iotaledger/crypto.rs/pull/31/commits/5670f988e03a959b9d6942c3bbbd08c3b8cb3619 together with a warning in its documentation.

Can you do a final review and perhaps create the issue referencing this "hole" in the current implementation?

rootmos: This is definitely a "stinker area", I do suggest we ask someone else to do another clean room implementation and compare/replace the one I've written here. Even better if said implementation is done on a system with different endianess (both bit and byte).

Wollac: I did found this commit https://github.com/bip32JP/bip32JP.github.io/commit/360c05a6439e5c461bbe5e84c7567ec38eb4ac5f saying `test vectors should be NOT normalized. Too many people relying on misconception that user INPUT will be == normalized text. If your library fails on this input, it will fail in real world.`
Previous to this commit the mnemonics have all been normalized and the tests pass.
Having non-normalized passphrase and mnemonic makes a lot of sense when you want to test the mnemonic->seed functionality. However, I guess for entropy->mnemonics I guess the best we can do is normalize before comparing.

rootmos: Which is what the code is doing now.

nothingismagick:
This seems to me to be resolved.

Wollac (approved):

PR #34, opened 2020-12-17
Add SHA384

l1h3r:
# Description of change

Adds SHA384 for JSON web tokens used in identity.rs

## Type of change

Copied existing test structure using RFC test vectors and ones generated with python

## Change checklist


Wollac (commented):
Can the test vectors be moved into a different file? `sha.rs` has gotten so big, that it cannot be reviewed in GitHub. Maybe under `tests` similar to curl and blake?
Also, ideally the test vectors should be in a separate file than the test code. That way, the code can be checked and diff'ed without having to look at megabytes of test data.

Wollac (approved):
lgtm

PR #41, opened 2021-01-10
Add AES-CBC

l1h3r:
# Description of change

Adds AES-CBC HMAC SHA2 algorithms defined in [rfc7518](https://tools.ietf.org/html/rfc7518#section-5.2) for JSON web tokens used in identity.rs

## Type of change

RFC test vectors

## Change checklist


nothingismagick:
You know, this sort of thing REALLY seems like it should be a procedure in stronghold tbh. But I understand why you need it here.

felsweg-iota (approved):
Looks good to me. Please let @Wollac approve as well.

Cargo.toml
```diff
@@ -87,6 +89,12 @@ curve25519-dalek = { version = "3", optional = true, default-features = false, f
 x25519-dalek = { version = "1.1", optional = true, default-features = false, features = [ "u64_backend" ] }
 zeroize = { version = "1.3", optional = true, default-features = false }
 
+
+[dependencies.subtle]
```
semenov-vladyslav: It'd be nice to keep all dependencies in the same style, eg. `subtle = { version = "2.4", optional = true, default-features = false }`

tests/ed25519.rs
```diff
@@ -124,7 +124,7 @@ fn test_eq_ord() -> crypto::Result<()> {
     assert!(pk == pk_eq);
     assert!(pk != pk_diff);
     assert!(pk > pk_diff);
-    assert!(!(pk > pk_eq));
+    assert!(pk <= pk_eq);
```
semenov-vladyslav: A couple lines above there's `assert!(pk == pk_eq);` already, it's safe to remove this line `assert!(pk <= pk_eq);`

src/ciphers/aes_cbc.rs
```diff
@@ -0,0 +1,192 @@
+// Copyright 2020 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+use aes_crate::{Aes128, Aes192, Aes256, BlockCipher, NewBlockCipher};
```
semenov-vladyslav, line 4: It seems that the `aes` crate was updated to version "0.8.1" in which `NewBlockCipher` was removed. Shall we also update or stick to the old version "0.7"?

Cargo.toml
```diff
@@ -25,6 +25,7 @@ name = "crypto"
 [features]
 default = [ ]
 std = [ ]
+aes-cbc = [ "aes-crate", "block-modes", "cipher", "hmac", "sha", "subtle" ]
```
semenov-vladyslav, line 28: Should "aes" feature imply "aes-cbc"?

semenov-vladyslav, line 28: AES-related features or doc-comment in `src/ciphers/traits.rs` for `Aead` trait seem a bit broken. The following commands fail (but should not):

```
cargo test --features="aes-cbc"
cargo test --features="aes-cbc aes"
```

semenov-vladyslav (approved):
Looks good enough. But there's a few minor things to improve upon.

Cargo.toml
Wollac: I think it probably should. Separating this seems to be a bit too fine granular IMHO

PhilippGackstatter: Specifying required features for cargo tests is a pain point, so I think one typically runs `cargo test --all-features`.

I think the alternative for these cases is

> cargo test --features="aes-cbc"

Make `aes-cbc` part of the `aes` feature completely, and don't expose it as a separate feature at all,

> cargo test --features="aes-cbc aes"

Activate the `random` feature whenever the `aead` feature is enabled, but that is bad style, imo, because we would only do that for tests.

I think the current solution is okay, and tests should be run with all features enabled.

src/ciphers/aes_cbc.rs
PhilippGackstatter: I tried upgrading, but this seems non-trivial. The `block_modes` crate was deprecated, as you pointed out, some traits were removed and instead `crypto-common` should now be used. Since the upgrade requires more than just replacing some imports, it seems beyond my expertise to do the upgrade. Personally, I'd be fine with getting the current version shipped and upgrading at a later point.

Cargo.toml
semenov-vladyslav: I think using `--all-features` is a bad idea as it hides problems with bad `#[cfg(feature...)]` attributes for imports or definitions which become unavailable if a certain feature is not enabled. This should be resolved in `Cargo.toml` with the right feature dependencies and with the right cfg attributes, and not by user enabling all features.

The failing doc-test requires `aes` feature which might not be enabled. Also, `aes` feature implies aes gcm, but not other modes (cbc, keywrap), which is also misleading.

I'd propose:
1. rename `aes` feature to `aes-gcm`;
2. create `aes = ["aes-cbc", "aes-gcm", "aes-kw"]` feature to enable all aes-related modes;
3. move doc-test in ciphers::traits into aes-gcm tests, or create a dedicated example, or somehow introduce `#[cfg(feature = "aes-gcm")]` to the doc-test.

PhilippGackstatter: I did the renaming as you suggested. In the process, I also renamed the `aes` module to `aes_gcm` so it is in line with the other modules `aes_cbc` and `aes_kw`. Note that this is a breaking change and requires a major version bump.

I feature gated the test you mentioned. It always succeeds if the required features are not enabled. No way around that, I think. The two test invocations you posted succeed now. Note that CI will still run the test because it uses `--all-features`, so it's fine overall.

Please have a look at the definition of the `aes` feature and if you agree with that.

Edit: If you are okay with the changes, please go ahead and merge. I don't have the necessary rights to do so.

tests/aead.rs
```diff
@@ -76,14 +77,35 @@ fn test_aead_all<A: Aead>(tvs: &[TestVector]) -> crypto::Result<()> {
 #[cfg(feature = "aes")]
```
semenov-vladyslav: Should be `#[cfg(feature = "aes-gcm")]`

tests/aead.rs
```diff
@@ -1,7 +1,7 @@
 // Copyright 2020 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
-#![cfg(any(feature = "aes", feature = "chacha"))]
+#![cfg(any(feature = "aes", feature = "chacha", feature = "aes-cbc"))]
```
semenov-vladyslav: Use `"aes-gcm"` feature here `#![cfg(any(feature = "aes-gcm", feature = "chacha", feature = "aes-cbc"))]`.

semenov-vladyslav (changes requested):
Please use `feature = "aes-gcm"` in tests/aead.rs

semenov-vladyslav (approved):

PR #45, opened 2021-01-11
X25519

rootmos:
Add the X25519 function to support key exchange using the Curve25519 curve.

My initial use-cases in [stronghold.rs](https://github.com/iotaledger/stronghold.rs) are:
* use an ephemeral keypair in the offline storage format (instead of the user provided secret directly): https://github.com/iotaledger/stronghold.rs/pull/123
* internal marshaling of secrets using a ["SealedBox"](https://doc.libsodium.org/public-key_cryptography/sealed_boxes) style construction: https://github.com/iotaledger/stronghold.rs/pull/124 (used here: https://github.com/iotaledger/stronghold.rs/blob/e6d06145f70f15bbc8e59f04d616ad5794bf253e/runtime/src/secret.rs#L77)


nothingismagick (changes requested):
Please update the README.md with a line in the table defining this algo.

nothingismagick:
Also, it would be nice to explain in the PR text why this is needed.

Wollac:
Yeah, I was also wondering about the same thing. What is the desired application of plain X25519?
Apart from that, it looks good to me.

rootmos:
Algorithm added to the algorithm table. Raison d'être added to PR description.

Wollac:
I am still confused with the description and the applications.
As far as I am concerned (and according to the linked RFC) X25519(k, U) computes the scalar multiplication [k]U on Curve25519.
This can then, of course, be used for Elliptic-curve Diffie–Hellman to derive a shared secret between two parties over insecure channels. I presume that this is exactly being used for a "SealedBox". But what exactly is the purpose in the snapshot context which, by definition, only involves one party?

This is why I think we should make the descriptions clearer to help with this:
- Maybe write something along the lines of "Add the X25519 function that performs a scalar multiplication on the Curve25519. This is for example used when implementing Diffie-Helman key exchange." (in changes)
- I don't think that overloading X25519 to use the base when U is omitted is good design. This is technically not how the X25519 function is defined, but just a short-cut in software which could cause confusion. `x25519_dalek` has `PublicKey::from` or `diffie_hellman` exactly for those different purposes of X25519 which is much more readable. I think wrappers like this would make the usage much easier and clearer. Or what have been the considerations against using them?

nothingismagick (approved):

nothingismagick:
@Wollac - see this PR for the justification: https://github.com/iotaledger/stronghold.rs/pull/123

Wollac:
OK, fair enough if the danger of the symmetric encryption leaking the key is a consideration, I understand.

However, my documentation and design concerns still stand.

Wollac:
In addition, it might be worth considering moving this ECDHE key generation used for the snapshots into `crypto.rs` (or even inside X25519). For example, the way how the symmetric cypher (nonce) is computed is not trivial, highly security relevant and requires some standardization, i.e. agreement on the hash function.
Furthermore, this would simplify implementations/applications of forward secrecy relying on ECDHE in the future.

_Update: Thinking about this... wouldn't such wrapper even remove the need for the public "overloaded" X25199?_

l1h3r:
@Wollac would love another review on this - I've refactored into a small wrapper around the `x25519_dalek` types. for now this is minimal and doesn't include the key generation behaviour from stronghold.rs, but it removes the need for the "overloaded" x25519 function and provides a base to build on

src/x25519.rs
```diff
@@ -0,0 +1,83 @@
...
+// RFC7748   Bernstein's 2006 paper
+// scalar/u  secret/public
+// X25519    Curve25519
+
+use core::convert::TryInto;
+
+pub const PUBLIC_KEY_LEN: usize = 32;
+pub const SECRET_KEY_LEN: usize = 32;
+
+/// An X25519 Shared Secret - the result of a Diffie-Hellman key exchange.
```
Wollac, line 17:
```suggestion
/// An X25519 Shared Secret - the result of a Diffie-Hellman key exchange.
///
/// Each party computes this from an (ephemeral) [`SecretKey`] and their counterparty's [`PublicKey`].
```

src/x25519.rs
```diff
@@ -0,0 +1,83 @@
...
+pub type SharedSecret = x25519_dalek::SharedSecret;
+
+/// An X25519 Public Key
+pub struct PublicKey(x25519_dalek::PublicKey);
+
+impl PublicKey {
+    /// Create a new [`PublicKey`] from bytes.
```
Wollac:
```suggestion
    /// Load a [`PublicKey`] from a slice of bytes.
```

src/x25519.rs
```diff
@@ -0,0 +1,83 @@
...
+    #[cfg(feature = "random")]
+    pub fn generate() -> crate::Result<Self> {
+        let mut bytes: [u8; SECRET_KEY_LEN] = [0; SECRET_KEY_LEN];
+
+        crate::rand::fill(&mut bytes[..])?;
+
+        Self::from_bytes(&bytes[..])
+    }
+
+    /// Create a new [`SecretKey`] from bytes.
```
Wollac:
```suggestion
    /// Load a [`SecretKey`] from a slice of bytes.
```

.changes/x25519.md
```diff
@@ -0,0 +1,6 @@
+---
+"iota-crypto": minor
+---
+
+Add the X25519 function (aka Curve25519) to add support for public key
+derivation and Diffie-Hellman exchanges using the Curve25519 curve.
```
Wollac:
```suggestion
Add X25519 Diffie-Hellman key exchange and public key derivation using the Curve25519 curve.
```

src/x25519.rs
```diff
@@ -0,0 +1,83 @@
...
+    pub fn to_bytes(&self) -> [u8; SECRET_KEY_LEN] {
+        self.0.to_bytes()
+    }
+
+    /// Returns the [`PublicKey`] which corresponds to this [`SecretKey`].
+    pub fn public_key(&self) -> PublicKey {
+        PublicKey((&self.0).into())
+    }
+
+    /// Computes a Diffie-Hellman [`SharedSecret`] with the given [`PublicKey`].
```
Wollac:
```suggestion
    /// Computes the Diffie-Hellman [`SharedSecret`] from the [`SecretKey`] and the given [`PublicKey`].
```

Wollac (approved):
Thanks, looks good to me!

I've just proposed a few minor improvements to the comments. Maybe consider whether they match the usual documentation style and whether they could help.

One additional remark: With this version we no longer have the bare, byte-oriented X25519 function, exactly as specified in RFC7748. I am perfectly find with leaving it out. I might even favor this, as I think that the DH API now is safer, more readable (and maybe even a bit faster). However, this could be a potential compromise with @rootmos previous approach.

nothingismagick:
@l1h3r - can we keep both interfaces? The one that Gustav wrote (so that we can use that in Stronghold) and your more contemporary version?

l1h3r:
> @l1h3r - can we keep both interfaces? The one that Gustav wrote (so that we can use that in Stronghold) and your more contemporary version?

This interface supports the same patterns that were being used in both stronghold.rs PRs - [Here's an example](https://github.com/l1h3r/stronghold.rs/commit/8906ab8745b005701844ffe2a700edf6590c13a3) that updates https://github.com/iotaledger/crypto.rs/pull/45 to use the new interface.

I'm fine keeping the previous version around until @rootmos says the new interface is good for stronghold.rs

lucasfernog (approved):

PR #52, opened 2021-02-16
Updatable Digest

### LINKED ISSUE #17: BLAKE2b: Provide an implementation of Digest
Currently, there is only `hash` computing the BLAKE2b hash of variable output sizes. This might not be sufficient in some scenarios and there should be BLAKE2b-256 implementation of `Digest`.

l1h3r:
# Description of change

Exports the `Digest` trait and `Output` type when using the `blake2b` and/or `sha` features.

Also adds a `Blake2b256` type and exports the `Sha{256,384,512}` types to use with the `Digest` trait

**This is a breaking change because it moves the `blake2b` module to `hashes/blake2b`**

## Type of change

Existing tests pass and tested in [identity.rs](https://github.com/iotaledger/identity.rs)

## Change checklist


nothingismagick (dismissed):

nothingismagick (changes requested):
this needs a .changes file

Wollac:
Actually, looking at this, the old `hash` function looks rather error-prone to me:
- It uses the length of the output buffer to instantiate completely different hash functions
- Having a Blake2b256 version statically always gives you the correct length

What do you guys think about adding another Blake2b512 (is that really needed anywhere?) and removing the `hash` while we are currently doing a breaking change anyway?

l1h3r:
> What do you guys think about adding another Blake2b512 (is that really needed anywhere?) and removing the `hash` while we are currently doing a breaking change anyway?

I don't think a `Blake2b512` type is needed anywhere but don't mind adding; I'm all for removing the `hash` function - @thibault-martinez ?

thibault-martinez:
@l1h3r If I remember well, `Digest` is automatically implemented because you implement all the other traits, right ? If that's the case and we do have a `digest` method, I don't know why we would keep the previous `hash` function.

l1h3r:
> @l1h3r If I remember well, `Digest` is automatically implemented because you implement all the other traits, right ? If that's the case and we do have a `digest` method, I don't know why we would keep the previous `hash` function.

Yeah, `Digest` is auto-implemented for us - I'll remove the old `hash` function 👍

src/hashes/mod.rs
```diff
@@ -1,8 +1,19 @@
...
 #[cfg(feature = "sha")]
 pub mod sha;
+
+#[cfg(feature = "digest")]
+#[doc(inline)]
+pub use digest::Digest;
+
+#[cfg(feature = "digest")]
+#[doc(inline)]
+pub use digest::Output;
```
thibault-martinez: Any reason not to merge them in a single `use` ?

thibault-martinez (approved):
LGTM

src/hashes/mod.rs
l1h3r: nope, just overlooked

Wollac (approved):
nice!

nothingismagick (approved):

PR #56, opened 2021-02-25
Added slip10 supported curves

melatron:
# Description of change

* Added slip10 supported curves

## Choose a type of change, and delete any options that are not relevant.

* All tests pass

## Change checklist

### DIFF
--- a/.changes/add-slip10-curves.md
+++ b/.changes/add-slip10-curves.md
@@ -2,4 +2,4 @@
 "iota-crypto": minor
 ---
 
-Add Slip10 supported curves.
+Add explicit Slip10 support for Ed25519 curve.
--- a/.changes/add-slip10-curves.md
+++ b/.changes/add-slip10-curves.md
@@ -1,0 +1,5 @@
+---
+"iota-crypto": minor
+---
+
+Add Slip10 supported curves.
--- a/src/slip10.rs
+++ b/src/slip10.rs
@@ -13,6 +13,21 @@ use alloc::vec::Vec;
 // https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki
 // https://en.bitcoin.it/wiki/BIP_0039
 
+pub enum Curve {
+    Ed25519,
+    Secp256k1,
+    Nist256P1,
+}
+
+impl Curve {
+    fn seedkey(&self) -> &[u8] {
+        match self {
+            Curve::Ed25519 => b"ed25519 seed",
+            Curve::Secp256k1 => b"Bitcoin seed",
+            Curve::Nist256P1 => b"Nist256p1 seed",
+        }
+    }
+}
 /// A seed is an arbitrary bytestring used to create the root of the tree.
 ///
 /// Several standards generate and/or restricts the size of the seed:
@@ -29,14 +44,14 @@ impl Seed {
         Self(bs.to_vec())
     }
 
-    pub fn to_master_key(&self, h_key: &[u8]) -> Key {
+    pub fn to_master_key(&self, curve: Curve) -> Key {
         let mut i = [0; 64];
-        HMAC_SHA512(&self.0, h_key, &mut i);
+        HMAC_SHA512(&self.0, curve.seedkey(), &mut i);
         Key(i)
     }
 
-    pub fn derive(&self, h_key: &[u8], chain: &Chain) -> crate::Result<Key> {
-        self.to_master_key(h_key).derive(chain)
+    pub fn derive(&self, curve: Curve, chain: &Chain) -> crate::Result<Key> {
+        self.to_master_key(curve).derive(chain)
     }
 }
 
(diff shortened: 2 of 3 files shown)


Wollac (changes requested):
Things get actually a much more tricky when non-Ed25519 curves are allowed.
So I am wondering what the motivation behind this PR is? Are the other curves really needed in some projects or is this PR "just" aiming for a complete SLIP-10 implementation?

Unfortunately, all the following issues are related to unmodified lines, so I am providing them as a list instead of PR comments.

The implementation in the PR is not correct, the following points need to be addressed:
- [Private key → private key derivation](https://github.com/satoshilabs/slips/blob/master/slip-0010.md#private-parent-key--private-child-key) in the function `child_key`:
  - non-hardened key derivation is only forbidden for Ed25519, for the other curves this must be allowed: I = HMAC-SHA512(Key = c<sub>par</sub>, Data = ser<sub>P</sub>(point(k<sub>par</sub>)) || ser<sub>32</sub>(i)). I.e. the actual curve must be used to compute the public key from the "key".
  - The `secret_key` is no longer just the first 32-bytes, i.e. parse<sub>256</sub>(I<sub>L</sub>), but actual curve arithmetic needs to be performed parse<sub>256</sub>(I<sub>L</sub>) + k<sub>par</sub> (mod n).
  - As such, it is now necessary to store the actual "key", i.e. secret key, and "chain code", i.e. the upper 32-bytes of the HMAC as part of the `key` instead of just having the HMAC.
  - Not all `secret_keys` are valid! If parse<sub>256</sub>(I<sub>L</sub>) ≥ n or parse<sub>256</sub>(I<sub>L</sub>) + k<sub>par</sub> (mod n) = 0 the key is invalid! In that case it should be retried with a different parameter. (As the chance of this happening is not very high, it might be acceptable to just return an error instead of re-trying.)
- [Public parent key → public child key](https://github.com/satoshilabs/slips/blob/master/slip-0010.md#public-parent-key--public-child-key):
  - Non Ed25519 curves support this type of derivation.
- There should be function to compute the public key (curve dependent) from a key.
- _All_ the [test vectors](https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki#test-vectors) from the spec should be added. With the current version, all of them would fail.

tensor-programming (approved):

tensor-programming:
> Things get actually a much more tricky when non-Ed25519 curves are allowed.
> So I am wondering what the motivation behind this PR is? Are the other curves really needed in some projects or is this PR "just" aiming for a complete SLIP-10 implementation?
> 
> Unfortunately, all the following issues are related to unmodified lines, so I am providing them as a list instead of PR comments.
> 
> The implementation in the PR is not correct, the following points need to be addressed:
> 
> * [Private key → private key derivation](https://github.com/satoshilabs/slips/blob/master/slip-0010.md#private-parent-key--private-child-key) in the function `child_key`:
>   
>   * non-hardened key derivation is only forbidden for Ed25519, for the other curves this must be allowed: I = HMAC-SHA512(Key = cpar, Data = serP(point(kpar)) || ser32(i)). I.e. the actual curve must be used to compute the public key from the "key".
>   * The `secret_key` is no longer just the first 32-bytes, i.e. parse256(IL), but actual curve arithmetic needs to be performed parse256(IL) + kpar (mod n).
>   * As such, it is now necessary to store the actual "key", i.e. secret key, and "chain code", i.e. the upper 32-bytes of the HMAC as part of the `key` instead of just having the HMAC.
>   * Not all `secret_keys` are valid! If parse256(IL) ≥ n or parse256(IL) + kpar (mod n) = 0 the key is invalid! In that case it should be retried with a different parameter. (As the chance of this happening is not very high, it might be acceptable to just return an error instead of re-trying.)
> * [Public parent key → public child key](https://github.com/satoshilabs/slips/blob/master/slip-0010.md#public-parent-key--public-child-key):
>   
>   * Non Ed25519 curves support this type of derivation.
> * There should be function to compute the public key (curve dependent) from a key.
> * _All_ the [test vectors](https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki#test-vectors) from the spec should be added. With the current version, all of them would fail.

Yeah, I actually agree with this.  Adding more curves for the sake of completeness seems like an easy way to overcomplicate things.  Unless they are specifically needed then there probably isn't a reason to do this.

tensor-programming (dismissed):
I agree with @Wollac. Adding more curves for the sake of completeness seems like an easy way to overcomplicate things.  Unless they are specifically needed for a project right now then there probably isn't a reason to do this. PR is likely unnecessary if the curves are not being used.

melatron:
The use case is to have the Ed25519 constant Curve directly in crypto.rs, otherwise we need to define it in every library that uses slip10. I can remove the other two if they are unnecessary and leave only Ed25519?

Wollac (approved):
I am happy with that change

.changes/add-slip10-curves.md
```diff
@@ -0,0 +1,5 @@
+---
+"iota-crypto": minor
+---
+
+Add Slip10 supported curves.
```
nothingismagick:
```suggestion
Add explicit Slip10 support for Ed25519 curve.
```

nothingismagick (approved):

PR #60, opened 2021-03-03
feat(ed25519): movev verify to PublicKey

nothingismagick:
# Description of change
This just places verify within the context of PublicKey.

## Type of change

Ran tests locally.

## Change checklist

### DIFF
--- a/src/signatures/ed25519.rs
+++ b/src/signatures/ed25519.rs
@@ -37,6 +37,10 @@ impl SecretKey {
 pub struct PublicKey(ed25519_zebra::VerificationKey);
 
 impl PublicKey {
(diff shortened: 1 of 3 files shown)


src/signatures/ed25519.rs
```diff
@@ -37,6 +37,10 @@ impl SecretKey {
 pub struct PublicKey(ed25519_zebra::VerificationKey);
 
 impl PublicKey {
+    pub fn verify(pk: &PublicKey, sig: &Signature, msg: &[u8]) -> bool {
```
thibault-martinez: The first parameter `pk: &PublicKey` should be `self` so that we can use it the same way zebra is using it `pk.verify(sig, msg)`

nothingismagick: aight

thibault-martinez (approved):

PR #62, opened 2021-03-03
Add Aead Trait

### LINKED ISSUE #43: Add Cipher trait
## Description

crypto.rs should provide a `Cipher` [trait](https://doc.rust-lang.org/rust-by-example/trait.html) to expose a consistent API and make it easier to use multiple algorithms with [generic types](https://doc.rust-lang.org/rust-by-example/generics.html).

## Motivation

This ensures all implementations have a consistent API and allows the following code:

```rust
use crate::ciphers::{aes::Aes256Gcm, chacha::XChaCha20Poly1305, traits::Cipher}

fn generic_encrypt<C: Cipher>(...) -> Result<<C as Cipher>::Tag> {
  C::encrypt(...)
}

let tag = generic_encrypt::<Aes256Gcm>(...)?;
// OR
let tag = generic_encrypt::<XChaCha20Poly1305>(...)?;
```

## Requirements

1. Add a `Cipher` trait, available at `crate::ciphers::traits::Cipher`
2. Provide a `Cipher` implementation for all relevant types in `crate::ciphers`
3. 🎉  Party 🎉 

## Open questions

An example of what this trait might look like is below.

Implementations need to specify:
  - key/nonce/tag lengths (as [typenum](https://docs.rs/typenum/1.12.0/typenum/) values, exposed via `consts`)
  - a custom error type (this example simply ignores the underlying implementation error)
  - a user-friendly name for error messages
  - `encrypt/decrypt` functions with explicitly sized arguments (where appropriate)

The trait would also provide easier-to-use `try_encrypt/try_decrypt` functions that accept rust [slices](https://doc.rust-lang.org/std/primitive.slice.html) as arguments.

Notes: 
  - The conversion in `try_generic_array` is simply a length check/pointer cast - no data is being copied
  - `decrypt` and `try_decrypt` return the length of the plaintext. This is to accommodate implementations that may apply padding.
- It may be worthwhile to add a function to calculate the padding of a plaintext buffer (`fn pad(plaintext: &[u8]) -> usize;`)
  - This is to further accommodate implementations with padding (a default implementation could be provided and just return `0`)

```rust
use aes::cipher::generic_array::{ArrayLength, GenericArray};
use core::convert::TryInto;

pub mod consts {
  pub use aes::cipher::generic_array::typenum::*;
}

pub type Key<T> = GenericArray<u8, T>;

pub type Nonce<T> = GenericArray<u8, T>;

pub type Tag<T> = GenericArray<u8, T>;

pub trait Cipher {
  type KeyLength: ArrayLength<u8>;
  type NonceLength: ArrayLength<u8>;
  type TagLength: ArrayLength<u8>;

  type Error;

  const NAME: &'static str;

  fn encrypt(
    key: &Key<Self::KeyLength>,
    iv: &Nonce<Self::NonceLength>,
    aad: &[u8],
    ptx: &[u8],
    ctx: &mut [u8],
    tag: &mut Tag<Self::TagLength>,
  ) -> Result<(), Self::Error>;

  fn decrypt(
    key: &Key<Self::KeyLength>,
    iv: &Nonce<Self::NonceLength>,
    aad: &[u8],
    tag: &Tag<Self::TagLength>,
    ctx: &[u8],
    ptx: &mut [u8],
  ) -> Result<usize, Self::Error>;

  fn try_encrypt(
    key: &[u8],
    iv: &[u8],
    aad: &[u8],
    ptx: &[u8],
    ctx: &mut [u8],
  ) -> crate::Result<Tag<Self::TagLength>> {
    let key: &Key<Self::KeyLength> = try_generic_array(key)?;
    let iv: &Nonce<Self::NonceLength> = try_generic_array(iv)?;
    let mut tag: Tag<Self::TagLength> = Default::default();

    Self::encrypt(key, iv, aad, ptx, ctx, &mut tag)
      .map_err(|_| crate::Error::CipherError { alg: Self::NAME })?;

    Ok(tag)
  }

  fn try_decrypt(
    key: &[u8],
    iv: &[u8],
    aad: &[u8],
    tag: &[u8],
    ctx: &[u8],
    ptx: &mut [u8],
  ) -> crate::Result<usize> {
    let key: &Key<Self::KeyLength> = try_generic_array(key)?;
    let iv: &Nonce<Self::NonceLength> = try_generic_array(iv)?;
    let tag: &Tag<Self::TagLength> = try_generic_array(tag)?;

    Self::decrypt(key, iv, aad, tag, ctx, ptx)
      .map_err(|_| crate::Error::CipherError { alg: Self::NAME })
  }
}

fn try_generic_array<T>(slice: &[u8]) -> crate::Result<&GenericArray<u8, T>>
where
  T: ArrayLength<u8>,
{
  if slice.len() == T::USIZE {
    Ok(slice.into())
  } else {
    Err(crate::Error::BufferSize {
      needs: T::USIZE,
      has: slice.len(),
    })
  }
}
```

Yes

l1h3r:
# Description of change

Adds an `Aead` trait to unify `AES-GCM` and `ChaCha` APIs (It will also be used for `AES-CBC` here: https://github.com/iotaledger/crypto.rs/pull/41).

Exports the following types to use with the trait:

- `Aes256Gcm`
- `XChaCha20Poly1305`

Closes https://github.com/iotaledger/crypto.rs/issues/43

## Type of change

Ran existing tests and added additional test vectors for `Aes256Gcm`

## Change checklist


nothingismagick (changes requested):
Since this is a breaking change, we would do well to document exactly how the new usage is different. At the very least we should put some documentation in the ciphers/traits.rs file

Otherwise this is a very forward thinking PR. :)

src/ciphers/traits.rs
```diff
@@ -0,0 +1,97 @@
...
+        0
+    }
+
+    fn encrypt(
+        key: &Key<Self::KeyLength>,
+        iv: &Nonce<Self::NonceLength>,
+        aad: &[u8],
+        ptx: &[u8],
+        ctx: &mut [u8],
+        tag: &mut Tag<Self::TagLength>,
```
thibault-martinez: This is probably just my lack of cipher knowledge but will all these parameters always make sense for all kind of ciphers ?

Wollac: I agree, this is an interface for GCM, not even AEAD, or even [AE](https://en.wikipedia.org/wiki/Authenticated_encryption) yet alone general encryption i.e. cypher.
In that sense, does it rely provide the sense of abstraction that we are aiming for?

Wollac: The very least, this should be called an AEAD trait. And use better names/descriptions for the parameters. Also nonce, and iv (initial vector) is not the same thing, so we should be consistent here.
Note that, `iv` and `aad` (which, I suppose, stands for associated data?) are optional, there are authenticating encryption schemes that do not rely on this.

Depending on the expected use-case, I would maybe propose to make this a GCM only trait and if necessary add a more general AE trait.

Wollac (commented):
In contrast to the title, this PR does various things:
- it adds new "ciphers" e.g. chacha20poly1305
- it moves tests around and
- it introduces an AEAD trait for GCM and Poly1305

It don't think these should be in a single PR as they have different aspects to it.

What is the driving force behind this PR:
- Do we have a project that relies on easily exchanging of CHACHA-POLY1305 and AES-GCM?
- Do we need authenticated encryption *with associated data* in this context? This is required, e.g. for packets that have to contain a public header and encrypted payload, but not for "general" encryption of private data?
- Do we need the option to explicitly provide a nonce or could this be replaced by a random generator?
- Should the trait be compatible with other AE(AD) schemes?

src/ciphers/traits.rs
l1h3r: Agreed that `Cipher` is not a good name but this is still an interface for AEAD, right? It was based on [`AeadInPlace`](https://docs.rs/aead/0.3.2/aead/trait.AeadInPlace.html) from the `aead` crate. I can go with the name `Aead`.

Will consistently use `nonce` and add more descriptive names 👍

l1h3r:
> In contrast to the title, this PR does various things:
> 
>     * it adds new "ciphers" e.g. chacha20poly1305
>     * it moves tests around and
>     * it introduces an AEAD trait for GCM and Poly1305
> 
> It don't think these should be in a single PR as they have different aspects to it.

Yeah, this PR is doing quite a bit. I will split it up into easier-to-consume PRs (tests/traits/impls)

> What is the driving force behind this PR:
> 
>     * Do we have a project that relies on easily exchanging of CHACHA-POLY1305 and AES-GCM?

My original motivation was to make things easier while working on JSON Web Tokens for identity.rs. Some examples of how it's being used are here:
https://github.com/iotaledger/identity.rs/blob/feat/libjose/libjose/src/jwe/encoder.rs#L342
https://github.com/iotaledger/identity.rs/blob/feat/libjose/libjose/src/jwe/encoder.rs#L628

I was planning to add AES-CBC which would also implement this trait:
https://github.com/iotaledger/crypto.rs/blob/libjose/src/aes_cbc.rs

>     * Do we need authenticated encryption _with associated data_ in this context? This is required, e.g. for packets that have to contain a public header and encrypted payload, but not for "general" encryption of private data?

Yeah, it will be used for identity-related communication - [DIDComm](https://identity.foundation/didcomm-messaging/spec/)

>     * Do we need the option to explicitly provide a nonce or could this be replaced by a random generator?

Generating one is fine for my needs but I'm not sure if that makes it too limiting for others

>     * Should the trait be compatible with other AE(AD) schemes?

Yes (to some extent I suppose). Personally I just want to add AES-CBC; I'm not sure what others would be needed

l1h3r:
Reduced this PR down to just the addition of the new trait:

- [Separated the test structures changes](https://github.com/iotaledger/crypto.rs/pull/63) (this PR is still based on that branch)
- Removed additional algorithms - will restore will later PRs
- Renamed `Cipher` -> `Aead`
- Added `Aead::random_nonce`. Enabled with the current `random` feature
- Consistently used `nonce`; overall naming improvements
- Docs and doctests

nothingismagick (approved):
OK, thanks for the revision and the migration notes.

Wollac (approved):
Thanks! It's much clearer now!

PR #71, opened 2021-03-13
Smooth out cipher API 

### LINKED ISSUE #70: Cleaner API for Ciphers
## Description

The current Cipher API is pretty complicated and inconsistent. The heavy use of the `generic_array` library makes type conversion tricky from standard types such as `Vectors` and `Slices`.  Also, when the cypher fails due to a type error, the error messages are not very descriptive; apparently it depends on the type. If you specify a slice without the appropriate length, in some cases you get an error describing that the length is wrong and in other cases you just get a generic panic.  

Example (from stronghold):

```rust
let mut tag = [0u8; Self::TAG_LEN];

...

XChaCha20Poly1305::encrypt(
            ...
            &mut tag
                .try_into()?,
            ...
           )?;
```    

Despite the tag being a mutable slice of size 16 and the `GenericArray` implementing a `TryInto` trait for the conversion (`&mut [u8; 16]`), the tag doesn't get updated by the encryption call. There is no compile time error for this and instead it becomes a runtime error. An appropriate implementation looks more like this:

```rust
let mut tag = vec![0u8; Self::TAG_LEN];

...

XChaCha20Poly1305::encrypt(
            ...
            tag.as_mut_slice()
                .try_into()?,
            ...
           )?;
```

This becomes confusing because the nonce can be written like so without causing the same kind of errors:

```rust
let mut nonce = [0u8; Self::NONCE_LEN];

...

 XChaCha20Poly1305::encrypt(
            ...
            &nonce.try_into()?,
            ...
        )?;
```

Also, the ordering of the parameters for the encryption and decryption functions are inconsistent. The Encryption function looks like this:
```rust
XChaCha20Poly1305::encrypt(
           key,
           nonce,
           associated_data,
           plaintext,
           ciphertext,
           tag,
        )
```

Where as the decryption function looks like this:

```rust
XChaCha20Poly1305::decrypt(
            key,
            nonce,
            associated_data,
            tag,
            ciphertext,
            plaintext
        )
```

The API would be much smoother if the position parameters stayed the same across the two different functions.  The functions `try_encrypt` and `try_decrypt` also contain the same parameter order and the `try_encrypt` function does not allow you to specify a tag. 

## Motivation

The point of this feature is to make the overall cipher API much more user friendly.  As it is right now, its unnecessarily troublesome and unpredictable. Documentation will help with some of this needless complexity, but the type conversion problems will still persist.  Presumably the `try` functions are meant to lessen this complexity, but the lack of a tag on `try_encrypt` makes using it inefficient. 

## Requirements

Write a list of what you want this feature to do.

1. Add more descriptive errors for failure states; mainly length errors and type errors.
2. Fix the ordering of the parameter lists on the methods and the lack of the `tag` parameter on the `try_encrypt` method (or make it clear in the docs that the `try_encrypt` function returns a new tag).
3. Add a better example of these methods in use then what exists in the tests folders at the moment.  Doc testing would be useful for something like this.  

Yes

tensor-programming:
# Description of change

These changes include slight modifications to the Cipher API to make it more user friendly and tweaks to the Buffer error type to give more information.  

addresses issue #70 

Is a breaking change because the parameter list was reordered for the `decrypt` and `try_decrypt` functions. 

## Type of change

- [x] Documentation Fix

Used the unit tests and some examples to test out the errors. 

## Change checklist

### DIFF
--- a/src/ciphers/macros.rs
+++ b/src/ciphers/macros.rs
@@ -24,6 +24,7 @@ macro_rules! impl_aead {
 
                 if plaintext.len() > ciphertext.len() {
                     return Err($crate::Error::BufferSize {
+                        name: "ciphertext",
                         needs: plaintext.len(),
                         has: ciphertext.len(),
                     });
@@ -31,8 +32,25 @@ macro_rules! impl_aead {
 
                 if tag.len() != Self::TAG_LENGTH {
                     return Err($crate::Error::BufferSize {
+                        name: "tag",
                         needs: Self::TAG_LENGTH,
                         has: tag.len(),
+                    });
+                }
+
+                if key.len() != Self::KEY_LENGTH {
+                    return Err($crate::Error::BufferSize {
+                        name: "key",
+                        needs: Self::KEY_LENGTH,
+                        has: key.len(),
+                    });
+                }
+
+                if nonce.len() != Self::NONCE_LENGTH {
+                    return Err($crate::Error::BufferSize {
+                        name: "nonce",
+                        needs: Self::NONCE_LENGTH,
+                        has: nonce.len(),
                     });
                 }
 
@@ -51,16 +69,41 @@ macro_rules! impl_aead {
                 key: &$crate::ciphers::traits::Key<Self>,
                 nonce: &$crate::ciphers::traits::Nonce<Self>,
                 associated_data: &[u8],
-                tag: &$crate::ciphers::traits::Tag<Self>,
-                ciphertext: &[u8],
                 plaintext: &mut [u8],
+                ciphertext: &[u8],
+                tag: &$crate::ciphers::traits::Tag<Self>,
             ) -> crate::Result<usize> {
                 use aead::{AeadInPlace, NewAead};
 
                 if ciphertext.len() > plaintext.len() {
                     return Err($crate::Error::BufferSize {
+                        name: "plaintext",
                         needs: ciphertext.len(),
                         has: plaintext.len(),
+                    });
+                }
+
+                if tag.len() != Self::TAG_LENGTH {
+                    return Err($crate::Error::BufferSize {
+                        name: "tag",
+                        needs: Self::TAG_LENGTH,
+                        has: tag.len(),
+                    });
+                }
+
+                if key.len() != Self::KEY_LENGTH {
+                    return Err($crate::Error::BufferSize {
+                        name: "key",
+                        needs: Self::KEY_LENGTH,
+                        has: key.len(),
+                    });
+                }
+
+                if nonce.len() != Self::NONCE_LENGTH {
+                    return Err($crate::Error::BufferSize {
+                        name: "nonce",
+                        needs: Self::NONCE_LENGTH,
+                        has: nonce.len(),
                     });
                 }
 
--- a/src/ciphers/macros.rs
+++ b/src/ciphers/macros.rs
@@ -12,6 +12,8 @@ macro_rules! impl_aead {
 
             const NAME: &'static str = $name;
 
+            /// Warning: type conversions on the tag type can be tricky. instead of `&mut tag.try_into().unwrap()` use
+            /// `(&mut tag).try_into().unwrap()`
             fn encrypt(
                 key: &$crate::ciphers::traits::Key<Self>,
                 nonce: &$crate::ciphers::traits::Nonce<Self>,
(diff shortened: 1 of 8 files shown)


nothingismagick (approved):
Makes sense to me.

nothingismagick (changes requested):
Oops. Two things here: We should document HOW the changes are breaking - even though its quite obvious if you look at the diff, and do that inside of the changefile. Thankyou!

src/ciphers/macros.rs
```diff
@@ -51,19 +69,44 @@ macro_rules! impl_aead {
...
 
                 if ciphertext.len() > plaintext.len() {
                     return Err($crate::Error::BufferSize {
+                        name: "plaintext",
                         needs: ciphertext.len(),
                         has: plaintext.len(),
                     });
                 }
 
+                if tag.len() != Self::TAG_LENGTH {
```
l1h3r: These length checks shouldn't be necessary when using `GenericArray`? This should be statically asserted.

I think the issue in your example from stronghold is because of this: https://docs.rs/generic-array/0.14.4/src/generic_array/lib.rs.html#172-174

which has the following: `assert_eq!(slice.len(), N::USIZE);`

This is called through the blanket implementation of `TryFrom`: https://doc.rust-lang.org/src/core/convert/mod.rs.html#589

> but the lack of a tag on try_encrypt makes using it inefficient.

Can `try_encrypt` just match the API of `encrypt`?

l1h3r (changes requested):

Wollac (approved):
I don't see any general or cryptographic issues with that :+1:

src/ciphers/macros.rs
tensor-programming: The length checks are necessary because for whatever reason, the assertions don't come through on the length panics in cases where you are doing type conversions.  And as I mentioned in the issue, there are some very strange issues which result from the type conversions where in lengths that should be supported in the `generic_array` type are seemingly not.  

And for the sake of the API, shouldn't there be some internal consistency regarding the error messages?  For instance, right now if I put in the wrong sized tag I get an error from crypto.rs but presuming the assertions are working, if I put in an wrong sized key, ill get that assertion from the panic.  From an aesthetic point of view, it would be nicer to just have a consistent set of descriptive errors. 

I did try matching `try_encrypt` and `encrypt` but there were issues with how the type conversion plays with mutability and returning a reference.  It is possible if we make it so that `try_encrypt` doesn't return the tag however.  Is that something that everyone would be ok with, because I can make the change if that's the case?  @l1h3r

l1h3r: I completely agree with you on the improvements to the API and I'm all for changing `try_encrypt` so it takes a `&mut tag` argument. 

My concern seeing the length check added is because `GenericArray` is a supporting type for most crypto here and an issue with compile-time lengths not being asserted sounds like a more serious bug?

l1h3r: aha

this conversion is no good
```
&mut tag.try_into().unwrap()
```

this one is good
```
(&mut tag).try_into().unwrap()
```

tensor-programming: Ahh well there you go.  I suspected that it was less a bug with the library and more of a rust specific idiosyncrasy.

If you want, I can remove all of these checks?  I know if you use the `try_*` functions you will get the errors from the `try_generic_array` calls where as the specific `encrypt/decrypt` functions would give you the panics from `generic_array`.

tensor-programming:
Wow, very strange, can't believe the tests failed on the bip39 `test_wordlist_codec_different_data_different_encodings` test because the corrupt RNG seemed to come back the same.

src/ciphers/macros.rs
l1h3r: If the `try_*` functions match the other signatures (with your ordering fixes) I think the overall API will be a bit clearer; definitely more consistent. It makes it much easier to move from `encrypt` to `try_encrypt`.

> If you want, I can remove all of these checks?

I would remove the checks - I still don't think they are needed, but maybe add a small comment about this unexpected rust behaviour?

tensor-programming: Well, ok, Ill remove the checks.  In which case, it makes the most sense to tell people that they should likely use the `try_*` functions as their first priority.   I already matched the API for the functions by adding the tag to the `try_encrypt` function.

l1h3r (approved):
🚀

nothingismagick (approved):

PR #79, opened 2021-03-24
Add x448 key exchange

l1h3r:
# Description of change

Adds x448 key exchange for JSON web tokens used in identity.rs

## Type of change

RFC test vectors

## Change checklist


felsweg-iota (changes requested):
Looks good to me. Fix the merge conflicts.

olivereanderson:
> Looks good to me. Fix the merge conflicts.

Should be done now.

felsweg-iota (approved):

src/keys/x448.rs
```diff
@@ -0,0 +1,81 @@
...
+///
+/// Each party computes this from an (ephemeral) [`SecretKey`] and their counterparty's [`PublicKey`].
+pub type SharedSecret = x448_crate::SharedSecret;
+
+/// An X448 Public Key
+pub struct PublicKey(x448_crate::PublicKey);
+
+impl PublicKey {
+    /// Load a [`PublicKey`] from a slice of bytes.
+    pub fn from_bytes(bytes: &[u8]) -> crate::Result<Self> {
```
semenov-vladyslav, line 21: There should exist some crypto.rs-wide style-guide on how to implement such conversions: to/from array/slice own function vs trait method which can or can't fail. I'd propose the following rules:

- use standard traits as much as possible: AsRef, From/TryFrom, etc.
- use `_bytes` suffix if argument or return value is of type `[u8; N]` and use `_slice` suffix with the type is `&[u8]`
- use `try_` prefix with functions that can fail, ie. return `Result`

src/keys/x448.rs
```diff
@@ -0,0 +1,81 @@
...
+            })
+    }
+
+    /// Returns the [`PublicKey`] as an array of bytes.
+    pub fn to_bytes(&self) -> [u8; PUBLIC_KEY_LEN] {
+        *self.0.as_bytes()
+    }
+
+    /// Returns the [`PublicKey`] as a slice of bytes.
+    pub fn as_bytes(&self) -> &[u8] {
```
semenov-vladyslav, line 36: Why not `impl AsRef<[u8]>`? Or `impl AsRef<[u8; PUBLIC_KEY_LEN]>`?

src/keys/x448.rs
```diff
@@ -0,0 +1,81 @@
...
+impl SecretKey {
+    /// Generate a new random [`SecretKey`].
+    #[cfg(feature = "random")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "random")))]
+    pub fn generate() -> crate::Result<Self> {
+        let mut bytes: [u8; SECRET_KEY_LEN] = [0; SECRET_KEY_LEN];
+
+        crate::utils::rand::fill(&mut bytes[..])?;
+
+        Self::from_bytes(&bytes[..])
```
semenov-vladyslav, line 53: Why not `Ok(Self(bytes.into()))`?

tests/fixtures/x448.rs
```diff
@@ -0,0 +1,25 @@
...
+        shared: "ce3e4ff95a60dc6697da1db1d85e6afbdf79b50a2412d7546d5f239fe14fbaadeb445fc66a01b0779d98223961111e21766282f73dd96b6f",
+    },
+    TestVector {
+        secret_a: "203d494428b8399352665ddca42f9de8fef600908e0d461cb021f8c538345dd77c3e4806e25f46d3315c44e0a5b4371282dd2c8d5be3095f",
+        public_a: None,
+        secret_b: None,
+        public_b: "0fbcc2f993cd56d3305b0b7d9e55d4c1a8fb5dbb52f8e9a1e9b6201b165d015894e56c4d3570bee52fe205e28a78b91cdfbde71ce8d157db",
+        shared: "884a02576239ff7a2f2f63b2db6a9ff37047ac13568e1e30fe63c4a7ad1b3ee3a5700df34321d62077e63633c575c1c954514e99da7c179d",
+    },
+    // https://tools.ietf.org/html/rfc7748#section-6.1
```
semenov-vladyslav: section-6.2

src/keys/x448.rs
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    /// Load a [`SecretKey`] from a slice of bytes.
+    pub fn from_bytes(bytes: &[u8]) -> crate::Result<Self> {
+        let array: [u8; SECRET_KEY_LEN] = bytes.try_into().map_err(|_| crate::Error::ConvertError {
```
semenov-vladyslav, line 58: This temporary value of `array` might leak secret key bytes. A safer way would be to construct `Self` with zero bytes as key and then do `self.0.as_mut().copy_from_slice(bytes)`. Unfortunately `x448::SecretKey` only exposes `From<[u8; 56]>` implementation which itself can leak secret key bytes.

src/keys/x448.rs
```diff
@@ -0,0 +1,81 @@
...
+
+/// An X448 Secret Key
+pub struct SecretKey(x448_crate::Secret);
+
+impl SecretKey {
+    /// Generate a new random [`SecretKey`].
+    #[cfg(feature = "random")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "random")))]
+    pub fn generate() -> crate::Result<Self> {
+        let mut bytes: [u8; SECRET_KEY_LEN] = [0; SECRET_KEY_LEN];
```
semenov-vladyslav, line 49: This `bytes` local variable will leak secret key bytes.

semenov-vladyslav (changes requested):
Three major issues that should be addressed throughout crypto.rs:

1. There should be one style-guide (is there already one?) for how to implement SecretKey/PublicKey types: naming and conversions to/from arrays/slices and common traits AsRef/From/TryFrom. Two options:
    - prepare such guide and review crypto.rs for compliance, or
    - claim it's not important and leave as is.
2. Leakage of secret bytes through temporary/local variables. Two options:
    - internally implementations use `Zeroize`, it should also be used for local variables, or
    - we can claim that such leaks are not threat enough and leave as is.
3. Trust level for third-party crates. Two options:
    - trust third-parties completely, always use latest version, no need to even run tests, or
    - do not trust third-party implementers, review not only PR to crypto.rs but also implementation code, use fixed versions of crates (eg. `x448 = "=0.6.0"`)

Wollac:
Seems like this PR is no longer needed. Please just re-open if it still is.

PR #80, opened 2021-03-24
Add secp256k1 signatures/key exchange

l1h3r:
# Description of change

Adds secp256k1 signatures and key exchange to use with JSON web tokens.

## Type of change

Tested through identity.rs

## Change checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,6 +38,7 @@ pbkdf = [ "pbkdf2" ]
 bip39 = [ "pbkdf", "hmac", "sha", "pbkdf", "unicode-normalization" ]
 bip39-en = [ "bip39" ]
 bip39-jp = [ "bip39" ]
+secp256k1 = [ "k256", "signature" ]
 slip10 = [ "hmac", "sha", "ed25519", "random", "serde" ]
 cipher = [ "aead", "generic-array" ]
 
@@ -122,6 +123,17 @@ chacha20poly1305 = { version = "0.7.1", optional = true }
   version = "1.0.123"
   optional = true
   features = [ "derive" ]
+
+  [dependencies.k256]
+  version = "0.7"
+  optional = true
+  default-features = false
+  features = ["ecdh", "ecdsa", "sha256"]
+
+  [dependencies.signature]
+  version = "1.2"
+  optional = true
+  default-features = false
 
 [dev-dependencies]
 hex = "0.4.2"
--- a/README.md
+++ b/README.md
@@ -39,6 +39,7 @@ To be included in this list an implementation must:
 | macs        | HMAC-SHA2-384      | [`hmac`](/src/macs/hmac.rs)         | [rfc][HMAC-RFC]            | `hmac`             | [official][HMAC-TEST]    | ★★★★☆ |
 | macs        | HMAC-SHA2-512      | [`hmac`](/src/macs/hmac.rs)         | [rfc][HMAC-RFC]            | `hmac`             | [official][HMAC-TEST]    | ★★★★☆ |
 | signatures  | Ed25519            | [`ed25519`](/src/signatures/ed25519.rs)        | [rfc (draft)][ED25519-RFC] | `ed25519-zebra`    | extended                 | ★★★★☆ |
+| signatures  | Secp256k1          | [`secp256k1`](/src/signatures/secp256k1rs)     | [spec][SECP256K1-SPEC]     | `k256`             | extended                 | ☆☆☆☆☆ |
 | utility    |  RANDOM                  | [`random`](/src/utils/rand.rs)              |                            | `getrandom`        | math                     | ★★★★★ |
 
 
@@ -143,3 +144,5 @@ Apache 2.0
 [BIP39-TEST]: https://github.com/bip32JP/bip32JP.github.io/blob/master/test_JP_BIP39.json
 
 [SLIP10-RFC]: https://github.com/satoshilabs/slips/blob/master/slip-0010.md
+
+[SECP256K1-SPEC]: https://www.secg.org/sec2-v2.pdf
--- a/src/error.rs
+++ b/src/error.rs
@@ -14,9 +14,11 @@ pub enum Error {
         needs: usize,
         has: usize,
     },
-    ///  Cipher Error
+    /// Cipher Error
     CipherError { alg: &'static str },
-    /// Convertion Error
+    /// Signature Error
+    SignatureError { alg: &'static str },
+    /// Conversion Error
     ConvertError { from: &'static str, to: &'static str },
     /// Private Key Error
     PrivateKeyError,
@@ -36,6 +38,7 @@ impl Display for Error {
                 write!(f, "{} buffer needs {} bytes, but it only has {}", name, needs, has)
             }
             Error::CipherError { alg } => write!(f, "error in algorithm {}", alg),
+            Error::SignatureError { alg } => write!(f, "error in signature algorithm {}", alg),
             Error::ConvertError { from, to } => write!(f, "failed to convert {} to {}", from, to),
             Error::PrivateKeyError => write!(f, "Failed to generate private key."),
             Error::InvalidArgumentError { alg, expected } => write!(f, "{} expects {}", alg, expected),
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -15,6 +15,12 @@ macro_rules! assert_buffer {
     }};
 }
 
+macro_rules! assert_buffer_eq {
+    ($length:expr, $target:expr, $name:expr) => {{
+        assert_buffer!($length, $target, $length == $target, $name)
+    }};
+}
+
 macro_rules! assert_buffer_lte {
     ($length:expr, $maximum:expr, $name:expr) => {{
         assert_buffer!($length, $maximum, $length <= $maximum, $name)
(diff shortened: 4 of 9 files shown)


felsweg-iota (approved):
Looks good to me. Please let @Wollac also approve.

Cargo.toml
```diff
@@ -123,6 +124,17 @@ chacha20poly1305 = { version = "0.7.1", optional = true }
   optional = true
   features = [ "derive" ]
 
+  [dependencies.k256]
```
semenov-vladyslav, line 127: Most other dependencies are in one-line form: `crate = { version = "1.1",  optional = true }`. Let's keep that style

README.md
```diff
@@ -39,6 +39,7 @@ To be included in this list an implementation must:
 | macs        | HMAC-SHA2-384      | [`hmac`](/src/macs/hmac.rs)         | [rfc][HMAC-RFC]            | `hmac`             | [official][HMAC-TEST]    | ★★★★☆ |
 | macs        | HMAC-SHA2-512      | [`hmac`](/src/macs/hmac.rs)         | [rfc][HMAC-RFC]            | `hmac`             | [official][HMAC-TEST]    | ★★★★☆ |
 | signatures  | Ed25519            | [`ed25519`](/src/signatures/ed25519.rs)        | [rfc (draft)][ED25519-RFC] | `ed25519-zebra`    | extended                 | ★★★★☆ |
+| signatures  | Secp256k1          | [`secp256k1`](/src/signatures/secp256k1rs)     | [spec][SECP256K1-SPEC]     | `k256`             | extended                 | ☆☆☆☆☆ |
```
semenov-vladyslav, line 42: Rating is confusing. What does no stars mean?

src/error.rs
```diff
@@ -14,9 +14,11 @@ pub enum Error {
         needs: usize,
         has: usize,
     },
-    ///  Cipher Error
+    /// Cipher Error
     CipherError { alg: &'static str },
-    /// Convertion Error
+    /// Signature Error
+    SignatureError { alg: &'static str },
```
semenov-vladyslav, line 20: `SignatureError` is overloaded here: it either means signature generation failed or signature verification failed. I'm not sure what to do in the first case - maybe try to generate it one more time (or just panic, it's a rare and hardly recoverable event). In the second case it's very much clear that the message signature is broken and we can't trust it. I'd propose to at least split `SignatureGenerationError` and `SignatureVerificationError` or something like that. Or extend `SignatureError` with an additional field to specify what exactly failed.

src/macros.rs
```diff
@@ -15,6 +15,12 @@ macro_rules! assert_buffer {
     }};
 }
 
+macro_rules! assert_buffer_eq {
+    ($length:expr, $target:expr, $name:expr) => {{
+        assert_buffer!($length, $target, $length == $target, $name)
```
semenov-vladyslav, line 20: This line asserts something about buffer lengths not buffers themselves. Would make sense to call it `assert_buffer_len!`.

src/signatures/secp256k1.rs
```diff
@@ -0,0 +1,168 @@
...
+/// Secp256k1 Public Key
+pub struct PublicKey(k256::PublicKey);
+
+// algorithm identifier used for errors
+const NAME: &str = "Secp256k1";
+
+impl PublicKey {
+    /// Load a [PublicKey] from a slice of bytes.
+    pub fn from_bytes(bytes: &[u8]) -> crate::Result<Self> {
+        assert_buffer_eq!(bytes.len(), UntaggedPointSize::<Secp256k1>::USIZE, "bytes");
```
semenov-vladyslav, line 32: Other SecretKey/PublicKey implementations in crypto.rs usually define SECRET_KEY_LEN/PUBLIC_KEY_LEN constants. Should it be the case here?

semenov-vladyslav: Also, it'd would be best to stick to the same pattern when implementing wrappers for secret/public keys.

src/signatures/secp256k1.rs
```diff
@@ -0,0 +1,168 @@
...
+            .decode()
+            .map(Self)
+            .map_err(|_| crate::Error::ConvertError {
+                from: "bytes",
+                to: "Secp256k1 Public Key",
+            })
+    }
+
+    /// Load a [PublicKey] from big-endian serialized coordinates.
+    pub fn from_coord(x: &[u8], y: &[u8]) -> crate::Result<Self> {
```
semenov-vladyslav, line 44: Do we need "coord" representation? Why not just keep those concatenated in one field / byte array?

src/signatures/secp256k1.rs
```diff
@@ -0,0 +1,168 @@
...
+            .try_into()
+            .map(Self)
+            .map_err(|_| crate::Error::ConvertError {
+                from: "Coordinates",
+                to: "Secp256k1 Public Key",
+            })
+    }
+
+    /// Return the [PublicKey] as an array of bytes.
+    pub fn to_bytes(&self) -> GenericArray<u8, UntaggedPointSize<Secp256k1>> {
```
semenov-vladyslav, line 65: Why not just `[u8; PUBLIC_KEY_LEN]`?

semenov-vladyslav: Rust now supports generic array legnths natively as in `fn f<const N: usize>() -> [u8; N]`. Shall we move closer to using this feature rather `GenericArray`?

src/signatures/secp256k1.rs
```diff
@@ -0,0 +1,168 @@
...
+                to: "Coordinates",
+            }),
+        }
+    }
+
+    /// Verify that the given `message` and `signature` are authentic.
+    pub fn verify(&self, message: &[u8], signature: &[u8]) -> crate::Result<()> {
+        signature::Signature::from_bytes(signature)
+            .and_then(|signature| Verifier::verify(self, message, &signature))
+            .map_err(|_| crate::Error::SignatureError { alg: NAME })
```
semenov-vladyslav, line 91: Authentication, verification, integrity check errors should stand out / be more alarming because they are caused by external causes, adversaries.

src/signatures/secp256k1.rs
```diff
@@ -0,0 +1,168 @@
...
+pub struct SecretKey(k256::SecretKey);
+
+impl SecretKey {
+    /// Generate a new random [SecretKey].
+    #[cfg(feature = "random")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "random")))]
+    pub fn generate() -> crate::Result<Self> {
+        let mut bytes: FieldBytes = Default::default();
+
+        crate::utils::rand::fill(&mut bytes[..])?;
```
semenov-vladyslav, line 111: `bytes` will leak secret key bytes. It should be zeroized. This issue is addressed in PR #79.

tests/secp256k1.rs
```diff
@@ -0,0 +1,44 @@
...
+    assert_eq!(pk.to_bytes(), sk.public_key().to_bytes());
+
+    let msg = utils::fresh::bytestring();
+    let mut sig = sk.sign(&msg).unwrap().as_ref().to_vec();
+
+    assert!(pk.verify(&msg, &sig).is_ok());
+    assert!(SecretKey::generate().unwrap().public_key().verify(&msg, &sig).is_err());
+
+    utils::corrupt(&mut sig);
+    assert!(pk.verify(&msg, &sig).is_err());
```
semenov-vladyslav, line 27: Corrupted `msg` should also fail.

src/signatures/secp256k1.rs
```diff
@@ -0,0 +1,168 @@
...
+        self.0.to_bytes()
+    }
+
+    /// Return the [PublicKey] which corresponds to this [SecretKey].
+    pub fn public_key(&self) -> PublicKey {
+        PublicKey(self.0.public_key())
+    }
+
+    /// Sign the given `message` and return a digital [signature][Signature].
+    pub fn sign(&self, message: &[u8]) -> crate::Result<Signature> {
```
semenov-vladyslav, line 137: Not clear how to handle `Err` case, it's not really under user's control

semenov-vladyslav (changes requested):
Looks good. Should mimic other secret/public key wrappers more closely.

tests/secp256k1.rs
```diff
@@ -0,0 +1,44 @@
...
+    let pkb = hex::decode("af81cf426b23df0f01c2f6acd86a7fa24cf48c1637c320e06f40dc315b597f6d6c6ccd209dfd4ea7ddb20c22b778c73dfbd44c2f7eeebed224878009fda33d16").unwrap();
+    let pk = PublicKey::from_bytes(&pkb).unwrap();
+
+    assert_eq!(pk.to_bytes(), sk.public_key().to_bytes());
+
+    let msg = utils::fresh::bytestring();
+    let mut sig = sk.sign(&msg).unwrap().as_ref().to_vec();
+
+    assert!(pk.verify(&msg, &sig).is_ok());
+    assert!(SecretKey::generate().unwrap().public_key().verify(&msg, &sig).is_err());
```
semenov-vladyslav, line 24: Error here, `generate` is not available. Add `#[cfg(feature = "random")]` to test declaration.

semenov-vladyslav (changes requested):
Failed test with `cargo test --features "secp256k1"`

src/signatures/secp256k1.rs
PhilippGackstatter: I'm not sure if we absolutely need it. I can't find a usage of the `from_coord` function, but we use the `to_coord` function for conversion into a `JWK`, here: https://github.com/iotaledger/identity.rs/blob/4287c9834de8dc42fd4ff0c0e8b3a2bc822d0ec4/libjose/src/jwk/key.rs#L487
For symmetry, I suppose it makes sense to keep both from and to functions, but if there's an alternative representation that achieves the same, I think we can also use that. Note that I'm not an expert on crypto math, so feel free to tell me what you think is best.

semenov-vladyslav: Right, I see. So, it's a standard way to represent secp256k1 points used by other implementations. In that case we should definitely keep it.

Wollac:
Seems like this PR is no longer needed. Please just re-open if it still is.

PR #104, opened 2021-07-27
Cosmetic enhancements for ed25519 and x25519

### LINKED ISSUE #103: Enhancements for ed25519 and x25519
## Description

Naming convention in ed25519 and x25519 modules is messy, some convenient methods/traits are not implemented, some functionality is missing.

## Motivation

1. In ed25519 there is no reason to specify *compressed* public key. *Compressed* representation is only used for serialization, uncompressed public key is not used. Why not just public key? Proposal: remove or deprecate names with `compressed` word and introduce corresponding names without it.

2. Unify `xxx_LENGTH` (`ed25519::PUBLIC_KEY_LENGTH`) and `xxx_LEN` (`x25519::PUBLIC_KEY_LEN`). `LEN` just looks ugly when other words are not shortened. Proposal: remove or deprecate `xxx_LEN` and introduce corresponding `xxx_LENGTH`.

3. There's no naming convention for conversions to/from bytes for PublicKey types (eg. `fn from_compressed_bytes(bs: &[u8]) -> Result<ed25519::PublicKey>`). Proposal to implement the following methods and traits:

    - `pub fn to_bytes(&self) -> [u8; PUBLIC_KEY_LENGTH]`
    - `pub fn as_slice(&self) -> &[u8]`
    - `pub fn from_bytes(bytes: [u8; PUBLIC_KEY_LENGTH]) -> Self` or `pub fn try_from_bytes(bytes: [u8; PUBLIC_KEY_LENGTH]) -> Result<Self>`
    - `impl Into<[u8; PUBLIC_KEY_LENGTH]> for &PublicKey`
    - `impl AsRef<[u8]> for PublicKey`
    - `impl From<[u8; PUBLIC_KEY_LENGTH]> for PublicKey` or `impl TryFrom<[u8; PUBLIC_KEY_LENGTH]> for PublicKey`

4. What is the endianness of to/from bytes conversions? Is it a "network" byte order or "native"? RFC 8032 states "All values are coded as octet strings, and integers are coded using little-endian convention", so LE must be the "network" byte order when converting x25519::SecretKey, PublicKeys, ed25519::Signature and x25519::SharedSecret as bytes as they are essentially big integers internally. ed25519::SecretKey is represented as seed bytes and not as a scalar that is derived from the seed, so the byte order here is natural. The old implementation is mixed up: ed25519::SecretKey used `to_le_bytes()`/`from_le_bytes()` which is not correct, and x25519::SecretKey bytes is scalar bytes in LE byte order used `to_bytes()`/`from_bytes()` with `le` infix missing. Also `le` functions suggest there are `be` variants which is confusing. Proposal: agree on "network" byte order which is LE (following RFC 8032), remove or deprecate `to_le_bytes`/`from_le_bytes`, use `to_bytes()`/`from_bytes()`.

5. Wrappers for PublicKey types do not currently implement common traits: `Copy`, `Clone`, `PartialEq`, `Eq`, `Hash`, `Debug`. Proposal: implement these traits. Rationale is trivial for each trait.

6. Generate SecretKey with prng. This allows SecretKeys to be deterministically generated from seed. Proposal: implement `generate_with<R>(rng: &mut R)` constructor.

7. Allow the use of the same secret key for ed25519 and x25519. Rationale: this is helpful as it simplifies key management: instead of two key pairs for signature and DH implementations can use only one `(ed25519::SecretKey,ed25519::PublicKey)` and convert it to x25519 keys when needed. Proposal: `impl From<ed25519::SecretKey/PublicKey> for x25519::SecretKey/PublicKey`.

8. Why `utils::rand::gen`s type is unrestricted? Proposal: Restrict to `Sized + Copy`. Rationale: Sized is obviously required to get size of T at compile time. Copy is needed by types that are copyable with memmove, ie. they are simple and don't contain pointers.

9. Deprecate or just remove/rename/breaking change? If deprecate, then since which version? Proposal: deprecate old names since "1.0.0" version.

10. What should be the convention for conversions for Copy types: `to_bytes(&self)` and `impl From<&PublicKey> for [u8; 32]` vs `to_bytes(self)` and `impl From<PublicKey> for [u8; 32]`? Proposal: follow [this clippy convention](https://rust-lang.github.io/rust-clippy/master/index.html#wrong_self_convention) and use the latter syntax.

## Requirements

Write a list of what you want this feature to do.

1. Address topics above.

## Open questions (optional)

Use this section to ask any questions that are related to the feature.

Yes.

semenov-vladyslav:
# Description of change

Unified naming convention, convenience methods, conversions, etc.

## Issue

This PR fixes issue #103

## Type of change

`cargo test --features "ed25519 x25519 rand random sha"`

## Change checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,7 +28,7 @@ std = [ ]
 aes-kw = [ "aes-crate" ]
 chacha = [ "chacha20poly1305", "cipher" ]
 ed25519 = [ "ed25519-zebra" ]
-x25519 = [ "x25519-dalek" ]
+x25519 = [ "x25519-dalek", "curve25519-dalek" ]
 random = [ "getrandom" ]
 aes = [ "aes-gcm", "cipher" ]
 blake2b = [ "blake2", "digest" ]
@@ -83,6 +83,7 @@ serde = { version = "1.0", optional = true, features = [ "derive" ] }
 sha3 = { version = "0.9", optional = true, default-features = false }
 tiny-keccak = { version = "2.0", optional = true, features = [ "keccak" ] }
 unicode-normalization = { version = "0.1", optional = true, default-features = false }
+curve25519-dalek = { version = "3", optional = true, default-features = false, features = [ "u64_backend" ] }
 x25519-dalek = { version = "1.1", optional = true, default-features = false, features = [ "u64_backend" ] }
 zeroize = { version = "1.3", optional = true, default-features = false }
 
--- a/.changes/ed25519_x25519_cleanup.md
+++ b/.changes/ed25519_x25519_cleanup.md
@@ -1,0 +1,7 @@
+---
+"iota-crypto": minor
+---
+
+* Unified naming convention in ed25519 and x25519 modules.
+* Added useful methods and standard traits implementations.
+* Added conversion of ed25519 keys to x25519 keys.
--- a/src/keys/slip10.rs
+++ b/src/keys/slip10.rs
@@ -62,11 +62,10 @@ pub type ChainCode = [u8; 32];
 pub struct Key([u8; 64]);
 
 impl Key {
-    pub fn secret_key(&self) -> crate::Result<SecretKey> {
+    pub fn secret_key(&self) -> SecretKey {
         let mut il = [0; 32];
         il.copy_from_slice(&self.0[..32]);
-        // TODO: this conversion should never fail
-        SecretKey::from_le_bytes(il)
+        SecretKey::from_bytes(il)
     }
 
     pub fn chain_code(&self) -> ChainCode {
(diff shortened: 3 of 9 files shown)


Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "iota-crypto"
-version = "0.6.0"
+version = "0.7.0"
```
nothingismagick: Please don't do this. We automatically bump 
Package versions with covector.

nothingismagick (changes requested):
Missing a .changes file. Please consult that directory for more information.

Cargo.toml
semenov-vladyslav: Sorry about that. Done

nothingismagick (approved):

PR #116, opened 2021-11-17
Add unrolled curl and batched curl

thibault-martinez:
## Change checklist


src/hashes/ternary/curl_p/batched/bct_curlp.rs
```diff
@@ -0,0 +1,125 @@
...
+    HASH_LENGTH,
+};
+
+pub(crate) struct BctCurlP {
+    state: BcTritArr<{ 3 * HASH_LENGTH }>,
+    scratch_pad: BcTritArr<{ 3 * HASH_LENGTH }>,
+}
+
+impl BctCurlP {
+    #[allow(clippy::assertions_on_constants)]
```
grtlr, line 22: There is only one `assert!` that this applies to, so we could move the `#[allow(...)]` directly above it.

src/hashes/ternary/curl_p/batched/mod.rs
```diff
@@ -0,0 +1,219 @@
...
+pub const BATCH_SIZE: usize = 8 * std::mem::size_of::<usize>();
+const HIGH_BITS: usize = usize::max_value();
+const NUM_ROUNDS: usize = 81;
+
+/// A hasher that can process several inputs at the same time in batches.
+///
+/// This hasher works by interleaving the trits of the inputs in each batch and hashing this
+/// interleaved representation. It is also able to fall back to the regular CurlP algorithm if
+/// required.
+pub struct BatchHasher<B: RawEncodingBuf> {
```
grtlr: Maybe we can come up with a more descriptive name here as it sounds very general, but it can actually only be used with `CurlP` at the moment. I'd be in favor of naming it `CurlPBatched` (or similar).

grtlr: It might make sense to think about streamlining the APIs between the regular and the batched function. For example, Regular `CurlP` is based on the sponge concept, whereas `BatchHasher` exposes `.hash_batched`.

src/hashes/ternary/curl_p/batched/bct_curlp.rs
thibault-martinez: Actually you can't put attributes on macros

src/hashes/ternary/curl_p/batched/mod.rs
```diff
@@ -0,0 +1,219 @@
...
+    type Item = TritBuf;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        let index = self.range.next()?;
+        Some(self.hasher.demux(index))
+    }
+}
+
+/// A helper iterator type for the output of the `hash_unbatched` method.
+struct UnbatchedHashes<'a, B: RawEncodingBuf> {
```
grtlr: Do we still need this?

grtlr (approved):

src/hashes/ternary/curl_p/batched/mod.rs
thibault-martinez: I can rename it but it's actually the internal `BctCurlP` that has a similar API to regular curl. The `BatchHasher` is just a proxy to make things easier, I don't really think there is a way to make it similar to regular Curl without hurting ease of use.

src/hashes/ternary/curl_p/batched/mod.rs
thibault-martinez: I think I would keep it, there is still a threshold (of number of provided hashes) where it's more efficient to not batch

src/hashes/ternary/curl_p/batched/mod.rs
grtlr: That makes sense. Should the name reflect that it uses Curl internally? Something like `CurlPBatchHasher`?

src/hashes/ternary/curl_p/batched/mod.rs
grtlr: I probably misunderstand, but wouldn't the user then use the unrolled version of Curl anyways?

src/hashes/ternary/curl_p/batched/mod.rs
thibault-martinez: Done

src/hashes/ternary/curl_p/batched/mod.rs
thibault-martinez: It was made like that for situations where you don't know in advance how many things you're going to hash, tbh I don't know if this is still a desired property or not

grtlr: I guess we can keep it for now and get rid of it once it becomes a burden to maintain.

thibault-martinez: I wouldn't create breaking changes in the future if not absolutely necessary. Maybe you are right, the only use case for this now is PoW and we always have 64 inputs ready for it, maybe we should remove

vuongDang:
Hey! great work.

For the tests shouldn't we also test against the inputs vectors of the "regular" CurlP? 
Also Wolfgang would like that we use the same test data as iota.go, would it be possible to run tests using this data as well? 
https://github.com/iotaledger/iota.go/blob/legacy/curl/testdata/curlp81.json

grtlr:
Thanks for the comment! I can look into incorporating these testcases.

src/hashes/ternary/curl_p/regular/u256.rs
```diff
@@ -0,0 +1,168 @@
...
+        self[(i / 64) % 4] |= 1 << (i % 64)
+    }
+
+    pub(super) fn shr_into(&mut self, x: &Self, shift: usize) -> &mut Self {
+        let offset = shift / 64;
+        let r = shift % 64;
+
+        if r == 0 {
+            for i in offset..4 {
+                self[(i - offset) % 4] |= x[i];
```
Wollac: Why do we have a mod 4 here? The loop makes sure that i - offset < 4, so it should not have any effect on the result. If this was added as part of some micro-optimizations to prevent compiler checks or similar, there should be a comment explaining this.

src/hashes/ternary/curl_p/regular/u256.rs
```diff
@@ -0,0 +1,168 @@
...
+
+    pub(super) fn set_bit(&mut self, i: usize) {
+        self[(i / 64) % 4] |= 1 << (i % 64)
+    }
+
+    pub(super) fn shr_into(&mut self, x: &Self, shift: usize) -> &mut Self {
+        let offset = shift / 64;
+        let r = shift % 64;
+
+        if r == 0 {
```
Wollac, line 39: This extra case is technically not necessary, shifting by 0 in the default case leads to the same result. So it was added because it turned out to be more performant in benchmarks? Otherwise consider removing this case.

src/hashes/ternary/curl_p/regular/u256.rs
```diff
@@ -0,0 +1,168 @@
...
+            3 => {
+                self[3] |= x[0] << l;
+            }
+            _ => {}
+        }
+
+        self
+    }
+
+    pub(super) fn norm243(&mut self) {
```
Wollac: It is by no means obvious what this function does. Consider adding a comment like "clears all the bits after the first 243" or at least make the cryptic and-mask a constant with an expressive name.

src/hashes/ternary/curl_p/regular/transform.rs
```diff
@@ -0,0 +1,126 @@
...
+
+        macro_rules! compute {
+            ($i: expr, $j: expr) => {
+                let tmp = batch_box(p[$i][$j], n[$i][$j], p2[$i][$j], n2[$i][$j]);
+                p[$i][$j] = tmp.0;
+                n[$i][$j] = tmp.1;
+            };
+        }
+
+        compute!(0, 0);
```
Wollac: Can we put all of this inside a double loop? I have no experience with automatic loop-unrolling in rust, but having a fixed integer range from 0 to 4 seems like a very easy case for the compiler. Then it would be much cleaner and we wouldn't even need the macro. Of course, if for whatever reason it does not get unrolled this could cause a significant slowdown and it is better to unroll manually like it is done here.

src/hashes/ternary/curl_p/regular/transform.rs
```diff
@@ -0,0 +1,126 @@
...
+
+    macro_rules! rotate {
+        ($p:expr, $p2:expr, $i:expr) => {
+            $p2[$i]
+                .shr_into(&$p[($i + offset) % 3], shift)
+                .shl_into(&$p[(($i + 1) + offset) % 3], 243 - shift);
+        };
+    }
+
+    rotate!(p, p2, 0);
```
Wollac: This could also potentially be fit into a loop if the unrolling works as desired. However, admittedly it wouldn't really clean the code up as much as above so it might be just worth keeping like this.

src/hashes/ternary/curl_p/regular/transform.rs
```diff
@@ -0,0 +1,126 @@
...
+    let mut n2 = <[U256; 3]>::default();
+
+    for i in 0..3 {
+        macro_rules! compute {
+            ($p:expr, $p2:expr, $j:expr, $m0:expr, $m1:expr, $m2:expr) => {
+                $p2[i][$j] = ($p[i][$j] & $m0) | ($p[(1 + i) % 3][$j] & $m1) | ($p[(2 + i) % 3][$j] & $m2);
+            };
+        }
+
+        compute!(p, p2, 0, M0, M1, M2);
```
Wollac, line 130: Also here one could put the `M`s inside an array and then write this as a loop and hope for the best. However, here I'd have serous doubts whether this would improve readability (and whether the compiler could unroll). So probably, just keep it like this.

src/hashes/ternary/curl_p/regular/transform.rs
```diff
@@ -0,0 +1,126 @@
...
+            state_rotation.offset = rotation / HASH_LENGTH;
+            state_rotation.shift = rotation % HASH_LENGTH;
+            rotation = (rotation * ROTATION_OFFSET) % STATE_SIZE;
+        }
+
+        state_rotations
+    };
+}
+
+pub(super) fn transform(p: &mut [U256; 3], n: &mut [U256; 3]) {
```
Wollac, line 59: These functions look even more like magic than they actually are :sweat_smile: Could we maybe at least at the function descriptions for `transform` and `reorder` from https://github.com/iotaledger/iota.go/blob/legacy/curl/transform.go?
Otherwise they are impossible to understand or reconstruct.

src/hashes/ternary/curl_p/batched/bct_curlp.rs
```diff
@@ -0,0 +1,125 @@
...
+    curl_p::batched::{
+        bct::{BcTrit, BcTritArr, BcTrits},
+        HIGH_BITS, NUM_ROUNDS,
+    },
+    HASH_LENGTH,
+};
+
+pub(crate) struct BctCurlP {
+    state: BcTritArr<{ 3 * HASH_LENGTH }>,
+    scratch_pad: BcTritArr<{ 3 * HASH_LENGTH }>,
```
Wollac: Scratchpad is a terrible name, can we rename this to something better that actually matches its purpose, like temp or state_copy or something.

src/hashes/ternary/curl_p/batched/bct_curlp.rs
```diff
@@ -0,0 +1,125 @@
...
+    pub(crate) fn reset(&mut self) {
+        self.state.fill(HIGH_BITS);
+    }
+
+    pub(crate) fn transform(&mut self) {
+        let mut scratch_pad_index = 0;
+
+        // All the unchecked accesses here are guaranteed to be safe by the assertion inside `new`.
+        for _round in 0..NUM_ROUNDS as usize {
+            self.scratch_pad.copy_from_slice(&self.state);
```
Wollac: Instead of copying the entire state in each round, one could just swap the pointers to the state and the copy which should be relevantly faster. However, I am not sure whether such a change would be out of scope of this PR.

src/hashes/ternary/curl_p/batched/bct_curlp.rs
```diff
@@ -0,0 +1,125 @@
...
+
+            let BcTrit(mut alpha, mut beta) = unsafe { *self.scratch_pad.get_unchecked(scratch_pad_index) };
```
Wollac: Can we please get rid of those legacy variable names `alpha` - `delta`. E.g. alpha should be something starting with `lo` and beta something with `hi`.

src/hashes/ternary/curl_p/batched/bct_curlp.rs
```diff
@@ -0,0 +1,125 @@
...
+
+            scratch_pad_index += 364;
+
+            let mut temp = unsafe { *self.scratch_pad.get_unchecked(scratch_pad_index) };
+
+            let delta = beta ^ temp.lo();
+
+            *unsafe { self.state.get_unchecked_mut(0) } = BcTrit(!(delta & alpha), delta | (alpha ^ temp.hi()));
```
Wollac: This is *the* key computation of Curl, the [S-box](https://en.wikipedia.org/wiki/S-box), could we move this into its own function exactly analog to the `batch_box` in the regular version? This would even remove the need for "delta" in `transform` and should make the code look much cleaner as now the same operation is performed multiple times.

src/hashes/ternary/curl_p/batched/bct_curlp.rs
```diff
@@ -0,0 +1,125 @@
...
+        let hash_count = trit_count / HASH_LENGTH;
+
+        for i in 0..hash_count {
+            unsafe { result.get_unchecked_mut(i * HASH_LENGTH..(i + 1) * HASH_LENGTH) }
+                .copy_from_slice(unsafe { self.state.get_unchecked(0..HASH_LENGTH) });
+
+            self.transform();
+        }
+
+        let last = trit_count - hash_count * HASH_LENGTH;
```
Wollac: The `regular` version only supports results that are multiples of `HASH_LENGTH`. I think the behavior should be consistent here.

src/hashes/ternary/curl_p/batched/bct_curlp.rs
```diff
@@ -0,0 +1,125 @@
...
+
+        unsafe { result.get_unchecked_mut(trit_count - last..trit_count) }
+            .copy_from_slice(unsafe { self.state.get_unchecked(0..last) });
+
+        if trit_count % HASH_LENGTH != 0 {
+            self.transform();
```
Wollac: A final `transform` is only needed when there is a successive call of `squeeze` later in the program. Especially, since most of the time squeeze will only be called once (for Chrysalis this will even always be the case, I think) this just performs an unnecessary and rather expensive transform operation. Is there a particular reason, why the `direction: SpongeDirection` approach from `regular` is not implemented here, as this fixes exactly this issue?

tests/batched_curlp.rs
```diff
@@ -0,0 +1,170 @@
...
+    for case in cases.iter() {
+        let object = case.as_object().unwrap();
+        let input = object["in"].as_str().unwrap();
+        let output = object["hash"].as_str().unwrap();
+        batched_curlp_broadcasted::<T1B1Buf, T1B1>(input, output);
+    }
+}
+
+#[test]
+fn batched_curl_p_broadcasted_t5b1_iota_go_json() {
```
Wollac: Probably I am missing something, but aren't `batched_curl_p_broadcasted_t1b1_iota_go_json` and `batched_curl_p_broadcasted_t5b1_iota_go_json` exactly identical?

Wollac (changes requested):

src/hashes/ternary/curl_p/batched/bct_curlp.rs
thibault-martinez: Are just `lo` and `hi` alright ?

Wollac: sure, if they are not clashing with anything

tests/batched_curlp.rs
grtlr: The underlying `TritBuf` should use a different encoding.

Wollac: I see, now it seems fixed.

src/hashes/ternary/curl_p/regular/u256.rs
thibault-martinez: Are you sure ? I tried removing it and tests failed

Wollac: I don't see why it shouldn't:
- If `r` is zero, `l` will be 64 and `x >> 0` should be exactly `x` and `x << 64` should always be zero.
- As such the `match` just becomes an unrolled version of the for loop.

But who am I to argue with test results :shrug: I mean we can certainly leave the for-loop in. It clearly doesn't impact the result to have it.

src/hashes/ternary/curl_p/batched/bct_curlp.rs
thibault-martinez: This was implemented before we implemented the unrolled version with direction. I can check to add it here as well.

src/hashes/ternary/curl_p/regular/u256.rs
pvdrz: That is how it was written in the go implementation. I agree that it is unnecessary in terms of code generation but at the same time I think it doesn't hurt having the immediate feedback of knowing that the index is not out of bounds by only reading this line instead of having to look up again what is `offset` and what are the possible `i` values and what-not.

src/hashes/ternary/curl_p/regular/u256.rs
pvdrz: Maybe bitwise operators behave differently from what we expect?

src/hashes/ternary/curl_p/regular/u256.rs
pvdrz: same name as the go impl. But I agree, the name is 0% clear

src/hashes/ternary/curl_p/regular/transform.rs
pvdrz: yes, the double loop should do the trick

src/hashes/ternary/curl_p/batched/bct_curlp.rs
pvdrz: I agree

src/hashes/ternary/curl_p/regular/transform.rs
pvdrz: it should be unrolled by the compiler automatically. Maybe we should dump this as a `cdylib` and take a look at the generated code to be sure. Want me to do it @thibault-martinez ?

src/hashes/ternary/curl_p/regular/u256.rs
Wollac: In Go this makes a difference as the compiler is not smart enough to detect that `(i-offset)` will never be out of bounds. After adding the `% 4` it becomes obvious to the compiler and saves us from rather costly runtime bound checks. I would guess that this would not be needed for the rust compiler...
However, if you think it improves readability then it is perfectly fine with me :+1:

src/hashes/ternary/curl_p/regular/u256.rs
Wollac: Apparently, maybe the integer overflow is undefined or something. Anyways, the special case should be a tiny bit faster anyways so let's just keep that.

Wollac (approved):

PR #127, opened 2022-02-07
Implement `try_from_slice()` for `ed25519::SecretKey` and `ed25519::PublicKey`

arnauorriols:
# Description of change

Added the conversion method `try_from_slice()` for the `ed25519::SecretKey` and `ed25519::PublicKey` types. The name of the method has been chosen following the apparent convention to call byte arrays `bytes` and byte slices `slice`.

The corresponding conversion traits have also been added, as well as the "missing" ones from existing methods of `ed25519::SecretKey`. 

I've following the same error handling of the existing fallible conversion method, `PublicKey::try_from_bytes()`, although it should be noted I'm not particularly fan of the error type used, as it loses the lower-level error type semantics.

## Type of change

- Enhancement (a non-breaking change which adds functionality)

`cargo test --all-features`

## Change checklist


Wollac (approved):

src/signatures/ed25519.rs
```diff
@@ -45,11 +45,37 @@ impl SecretKey {
...
     }
 }
 
+impl AsRef<[u8]> for SecretKey {
+    fn as_ref(&self) -> &[u8] {
+        self.as_slice()
+    }
+}
+
+impl From<[u8; PUBLIC_KEY_LENGTH]> for SecretKey {
```
semenov-vladyslav, line 66: It should be `SECRET_KEY_LENGTH`  here and a line below.

src/signatures/ed25519.rs
```diff
@@ -3,7 +3,7 @@
 
 use core::{
     cmp::Ordering,
```
semenov-vladyslav, line 5: Do we need `cmp::Ordering`? What for?

semenov-vladyslav (changes requested):
Missing .changes file. Typo in `From` instance for `SecretKey`.

PR #139, opened 2022-05-17
Derive traits for Curve, Key, Segment and Chain

Thoralf-M:
# Description of change

Derive some traits for Curve, Key, Segment and Chain so we can use it in wallet.rs

## Type of change

## Change checklist

### DIFF
--- a/src/keys/slip10.rs
+++ b/src/keys/slip10.rs
@@ -56,7 +56,7 @@ impl Seed {
 
 pub type ChainCode = [u8; 32];
 
-#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
+#[derive(Clone, Copy, Debug)]
 pub struct Key([u8; 64]);
 
 impl Key {
@@ -117,7 +117,7 @@ impl TryFrom<&[u8]> for Key {
     }
 }
 
-#[derive(Clone, Debug, Eq, Hash, PartialEq, Serialize, Deserialize)]
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq, Serialize, Deserialize)]
 pub struct Segment {
     hardened: bool,
     bs: [u8; 4],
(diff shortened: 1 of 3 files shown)


src/keys/slip10.rs
```diff
@@ -56,7 +56,7 @@ impl Seed {
 
 pub type ChainCode = [u8; 32];
 
-#[derive(Copy, Clone, Debug)]
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
```
semenov-vladyslav: Is it a secret key? If so then it should also derive `Zeroize`, equality should be constant time (eg. from subtle crate, `PartialEq` does not guarantee constant-time comparison), and I'm against `Hash` instances for secret keys as it might leak info about secret key.

src/keys/slip10.rs
```diff
@@ -56,7 +56,7 @@ impl Seed {
...
 pub struct Key([u8; 64]);
 
 impl Key {
```
semenov-vladyslav, line 63: `Key::secret_key` might leak secret key bytes via local variable `il` two lines below.

src/keys/slip10.rs
semenov-vladyslav: `Hash` can be implemented in a secure manner via some secure HKDF construction.

src/keys/slip10.rs
```diff
@@ -117,7 +117,7 @@ impl TryFrom<&[u8]> for Key {
     }
 }
 
-#[derive(Debug, Clone, Serialize, Deserialize)]
+#[derive(Clone, Debug, Eq, Hash, PartialEq, Serialize, Deserialize)]
```
semenov-vladyslav: Can also derive `Copy`

semenov-vladyslav (changes requested):
`Hash` and `PartialEq` default instances might be insecure for secret keys.

semenov-vladyslav (approved):
There's still an issue with secret bytes leakage via local variable. Unrelated to this PR.

semenov-vladyslav (approved):
Looks good

Wollac (changes requested):
Missing `.changes` file. Please add one; consult that directory for more information.

Everything else looks good to me.

Wollac (approved):

PR #155, opened 2022-08-12
Add aead_encrypt and aead_decrypt convenience functions

thibault-martinez:
## Change checklist

### DIFF
--- a/.changes/aead-helpers.md
+++ b/.changes/aead-helpers.md
@@ -1,0 +1,5 @@
+---
+"iota-crypto": minor
+---
+
+Add aead_encrypt and aead_decrypt convenience functions
--- a/src/ciphers/macros.rs
+++ b/src/ciphers/macros.rs
@@ -70,6 +70,9 @@ macro_rules! impl_aead {
             }
         }
 
+        /// A helper function to encrypt `plaintext` with `key`.
+        /// The return value is arbitrarily chosen as `nonce || tag || ciphertext` for historic reasons, mainly compatibility with out wallet libraries.
+        /// The nonce is randomly chosen.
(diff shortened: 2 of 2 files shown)


Wollac (commented):
Two remarks:
- While I am not aware of any particular specification and there are cases with a different order, I would say the "usual" order is `nonce || ciphertext || tag`. So should we also maybe use that order here?
- I absolutely agree, we should add such a util function. Especially since `associated_data` and `tag` is hardly ever needed explicitly. However, I see this a bit differently for the nonce. There are common scenarios where the nonce is not chosen randomly, e.g. all-zero for ephemeral keys or derived from a hash. Good arguments can be made, why the nonce should still be explicit (like e.g. in crypto_secretbox_easy and libsodium is usually a great example for API design).
However, nonce selection is security critical. So, when this better fits the use-case your are having in mind, let's leave it like this and hide the nonce.

thibault-martinez:
> * While I am not aware of any particular specification and there are cases with a different order, I would say the "usual" order is `nonce || ciphertext || tag`. So should we also maybe use that order here?

I kind of assumed that was the usual order as it has been chosen this way for the stronghold integration in the wallet. We obviously can't change it now. I would understand if choosing an arbitrary order was a no-go for an inclusion in crypto.rs

>   However, nonce selection is security critical. So, when this better fits the use-case your are having in mind, let's leave it like this and hide the nonce.

I am not fully certain what your conclusion is. Do you want me to make the nonce a parameter?

Wollac:
If that order is already in use let's stick with it :+1: 
Maybe you could add a comment describing the order and that the nonce is initialized randomly.

Wollac (approved):

thibault-martinez:
> Maybe you could add a comment describing the order and that the nonce is initialized randomly.

Done, let me know if this is fine.

PR #178, opened 2023-03-06
Bump dependencies

thibault-martinez:
## Change checklist

### DIFF
--- a/src/ciphers/aes_cbc.rs
+++ b/src/ciphers/aes_cbc.rs
@@ -1,27 +1,25 @@
-// Copyright 2020 IOTA Stiftung
+// Copyright 2023 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
 use core::{
     marker::PhantomData,
     num::NonZeroUsize,
-    ops::{Shl, Sub},
+    ops::{Shl, Shr},
 };
 
-use aes_crate::{Aes128, Aes192, Aes256, BlockCipher, NewBlockCipher};
-use block_modes::{block_padding::Pkcs7, BlockMode, Cbc};
+use aes_crate::cipher::{
+    block_padding::{Padding, Pkcs7},
+    BlockCipher, BlockDecryptMut, BlockEncryptMut, BlockSizeUser, InnerIvInit, KeyInit, KeySizeUser,
+};
+use aes_crate::{Aes128, Aes192, Aes256};
+use cbc_crate::{Decryptor as CbcDecryptor, Encryptor as CbcEncryptor};
 use generic_array::{
-    sequence::Split,
-    typenum::{Double, Unsigned, B1, U16, U24, U32},
+    typenum::{Unsigned, B1},
     ArrayLength,
 };
 use hmac_::{
-    digest::{
-        block_buffer::Eager,
-        core_api::{BlockSizeUser, BufferKindUser, CoreProxy, FixedOutputCore, UpdateCore},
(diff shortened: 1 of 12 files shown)


src/keys/age.rs
```diff
@@ -31,7 +31,7 @@ const MAC_BASE64_LEN: usize = 43;
 fn derive_wrap_key(password: &[u8], salt: &[u8; 16], work_factor: u8, wrap_key: &mut [u8; 32]) {
     // wrap key = scrypt(N = work factor, r = 8, p = 1, dkLen = 32,
     //     S = "age-encryption.org/v1/scrypt" || salt, P = passphrase)
-    let params = ScryptParams::new(work_factor, 8, 1).unwrap();
+    let params = ScryptParams::new(work_factor, 8, 1, 32).unwrap();
```
thibault-martinez, line 34: @semenov-vladyslav Just confirming, 32 is the right value?

semenov-vladyslav: Right, the derived from password via scrypt wrap key is 32-byte key used to protect file key via ChaCha20-Poly1305.

src/macs/hmac.rs
```diff
@@ -6,26 +6,26 @@
 #[cfg(feature = "sha")]
 #[cfg_attr(docsrs, doc(cfg(feature = "sha")))]
 pub fn HMAC_SHA256(data: &[u8], key: &[u8], mac: &mut [u8; 32]) {
-    use hmac_::Mac;
-    let mut m = hmac_::Hmac::<sha2::Sha256>::new_from_slice(key).unwrap();
+    use hmac::Mac;
```
semenov-vladyslav, line 9: I feel like we shouldn't use Mac trait here, it's going to change in the next version. Should probably use KeyInit and Update/Digest instead.

semenov-vladyslav (dismissed):
Last final update in hmac needed.

src/macs/hmac.rs
semenov-vladyslav: `Mac` trait in [digest-0.11](https://docs.rs/digest/0.11.0-pre/digest/trait.Mac.html) will not have `new` method. Right now there's no need to prepare for that.

semenov-vladyslav (approved):
Looks good.

src/ciphers/aes_cbc.rs
```diff
@@ -1,141 +1,81 @@
...
 };
 use sha2::{Sha256, Sha384, Sha512};
 use subtle::ConstantTimeEq;
+use zeroize::Zeroize;
 
 use crate::ciphers::traits::{Aead, Key, Nonce, Tag};
 
 /// AES-CBC using 128-bit key and HMAC SHA-256.
-pub type Aes128CbcHmac256 = AesCbc<Aes128, Sha256, U16, U16>;
+pub type Aes128CbcHmac256 = CbcHmac<Aes128, Pkcs7, Sha256>;
```
thibault-martinez, line 33: Out of curiosity, was `AesCbc` previously automatically using `Pkcs7`? Or why do we need to put it now?

semenov-vladyslav: Yes, for some reason, although there could be other paddings. Also, it was named AesCbc but it was possible to use ciphers other than AES. Having padding as generic argument just feels natural in the current implementation.

thibault-martinez: Thanks

PR #187, opened 2023-04-27
Secp256k1

semenov-vladyslav:
# Description of change

Please write a summary of your changes and why you made them.

## Type of change

- [ ] Documentation Fix

## Change checklist


tests/slip10.rs
```diff
@@ -21,35 +23,67 @@ mod test {
...
+    fn ed25519_test_vectors() -> Result<()> {
+        let tvs = include!("fixtures/slip10_ed25519.rs");
+
+        run_test_vectors(Curve::Ed25519, &tvs)
+    }
+
+    #[cfg(feature = "secp256k1")]
+    #[test]
+    fn secp256k1_test_vectors() -> Result<()> {
+        let tvs = include!("fixtures/slip10_secp256k1.rs");
```
Thoralf-M, line 85: Missing fixtures file?

src/signatures/secp256k1_ecdsa.rs
```diff
@@ -0,0 +1,198 @@
...
+}
+
+impl TryFrom<[u8; PUBLIC_KEY_LENGTH]> for PublicKey {
+    type Error = crate::Error;
+    fn try_from(bytes: [u8; PUBLIC_KEY_LENGTH]) -> crate::Result<Self> {
+        Self::try_from_bytes(&bytes)
+    }
+}
+
+// impl Hash for PublicKey {
```
Thoralf-M: Why commented?

tests/secp256k1_ecdsa.rs
```diff
@@ -0,0 +1,61 @@
+// Copyright 2021 IOTA Stiftung
```
Thoralf-M:
```suggestion
// Copyright 2023 IOTA Stiftung
```

README.md
```diff
@@ -42,6 +42,7 @@ To be included in this list an implementation must:
 | macs        | HMAC-SHA2-384      | [`hmac`](/src/macs/hmac.rs)         | [rfc][HMAC-RFC]            | `hmac`             | [official][HMAC-TEST]    | ★★★★☆ |
 | macs        | HMAC-SHA2-512      | [`hmac`](/src/macs/hmac.rs)         | [rfc][HMAC-RFC]            | `hmac`             | [official][HMAC-TEST]    | ★★★★☆ |
 | signatures  | Ed25519            | [`ed25519`](/src/signatures/ed25519.rs)        | [rfc (draft)][ED25519-RFC] | `ed25519-zebra`    | extended                 | ★★★★☆ |
+| signatures  | Secp256k1          | [`secp256k1`](/src/signatures/secp256k1rs)     | [spec][SECP256K1-SPEC]     | `k256`             | extended                 | ☆☆☆☆☆ |
```
thibault-martinez: You rate it 0 out of 5 :trollface: ?

src/macros.rs
```diff
@@ -15,6 +15,12 @@ macro_rules! assert_buffer {
     }};
 }
 
+macro_rules! assert_buffer_eq {
```
thibault-martinez: What is this for?

src/signatures/secp256k1_ecdsa.rs
```diff
@@ -0,0 +1,198 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+use core::convert::TryFrom;
+
+use zeroize::ZeroizeOnDrop;
+
+pub const SECRET_KEY_LENGTH: usize = 32;
+pub const PUBLIC_KEY_LENGTH: usize = 33;
+pub const SIGNATURE_LENGTH: usize = 65;
+pub const ADDRESS_LENGTH: usize = 20;
```
thibault-martinez: It's usually a good thing to have these as associated const of the types, so we don't have to import them when we need them.

tests/slip10.rs
semenov-vladyslav: Definitely. Sorry about that

src/signatures/secp256k1_ecdsa.rs
semenov-vladyslav: Not sure we need Hash for PublicKey. It's implemented for ed25519::PublicKey. Probably, better also impl Hash here too.

tests/secp256k1_ecdsa.rs
semenov-vladyslav: Done

README.md
semenov-vladyslav: Yeah, didn't like it) It's from older version of secp256k1 PR. Not sure how to properly rate it.

src/macros.rs
semenov-vladyslav: Not sure honestly. I kept it from the original secp256k1 PR. Better remove it, I think

src/signatures/secp256k1_ecdsa.rs
semenov-vladyslav: Good point. Those are not associated const for Ed25519. Shall we make them so there too?

thibault-martinez: I'd do it, yeah

src/macros.rs
thibault-martinez: Looks completely unused yeah

src/signatures/secp256k1_ecdsa.rs
semenov-vladyslav: I'd say we should keep global consts deprecated too. Otherwise we might break too much

semenov-vladyslav: What would be a good name for associated consts: SecretKey::LENGTH?

thibault-martinez: Yeah I think that's enough 👍🏻

src/signatures/secp256k1_ecdsa.rs
semenov-vladyslav: Ok, removed it. We'll add it later if it's needed. PublicKey types from k256, ecdsa, and elliptic_curve crates do not impl Hash.

README.md
semenov-vladyslav: Updated a bit.

src/macros.rs
semenov-vladyslav: Removed

Thoralf-M: Not yet?

src/signatures/secp256k1_ecdsa.rs
semenov-vladyslav: Done. Use XXX::LENGTH consts for secp256k1_ecdsa/ed25519 types, deprecated ed25519::XXX_LENGTH consts.

thibault-martinez: Thanks, maybe it deserves a little entry in the changelog?

src/macros.rs
thibault-martinez: Looks gone to me

src/signatures/secp256k1_ecdsa.rs
Thoralf-M: I think it's a useful thing to have, imagine a simple HashMap with public keys and their related accounts data (EVM)

tests/fixtures/slip10_ed25519.rs
```diff
@@ -30,11 +30,6 @@
                 chain_code: "8f6d87f93d750e0efccda017d662a1b31a266e4a6f5993b15f5c1f07f74dd5cc",
                 private_key: "30d1dc7e5fc04c31219ab25a27ae00b50f6fd66622f6e9c913253d6511d1e662",
             },
-            TestChain {
-                chain: Chain::from_u32_hardened(vec![0, 1, 2, 2]),
-                chain_code: "8f6d87f93d750e0efccda017d662a1b31a266e4a6f5993b15f5c1f07f74dd5cc",
-                private_key: "30d1dc7e5fc04c31219ab25a27ae00b50f6fd66622f6e9c913253d6511d1e662",
-            },
```
thibault-martinez, line 37: Why is this removed?

semenov-vladyslav: That was an exact duplicate of the test case above.

src/signatures/secp256k1_ecdsa.rs
semenov-vladyslav: Done

src/signatures/secp256k1_ecdsa.rs
semenov-vladyslav: Alright, implemented Hash for Address and PublicKey

Thoralf-M: Thanks!

src/keys/slip10.rs
```diff
@@ -44,100 +99,402 @@ impl Seed {
...
+    fn secret_bytes(&self) -> &[u8; 32] {
+        unsafe { &*(self.ext[1..33].as_ptr() as *const [u8; 32]) }
+    }
+
+    fn public_bytes(&self) -> &[u8; 33] {
+        unsafe { &*(self.ext[..33].as_ptr() as *const [u8; 33]) }
+    }
+
+    fn chain_code(&self) -> &[u8; 32] {
+        unsafe { &*(self.ext[33..].as_ptr() as *const [u8; 32]) }
```
Thoralf-M, line 342: Maybe also use consts for the 32, 33, 64, 65 numbers here?
`[1..33]` is the first byte a type byte or why is it 33?

semenov-vladyslav: SLIP-10 is designed for certain curves such as Secp256k1, NIST P-256, Ed25519. These consts are related to curve size, field modulus size, encoded public key size. KeyImpl is a private type, there's no need to name them. Maybe for ExtendedSecretKey and ExtendedPublicKey (extended_bytes, secret_bytes, public_bytes) it makes sense.

The ext[0] byte is there for encoding of a public key using SEC1 format. KeyImpl is used to derive both secret and public keys: for secret key ext[0] == 0 and ext[1..33] encodes a secret key, for public key ext[0..33] contains SEC1 encoding of a public key with ext[0] == 2 or ext[0] == 3. This is a way to implement polymorphism here. Otherwise there would be a copy-paste or some private trait.

Cargo.toml
```diff
@@ -111,6 +113,8 @@ zeroize = { version = "1.5", optional = true, default-features = false, features
 scrypt = { version = "0.11", optional = true, default-features = false }
 hkdf = { version = "0.12", optional = true, default-features = false }
 base64 = { version = "0.21", optional = true, default-features = false }
+k256 = { version = "0.13", optional = true, default-features = false, features = ["ecdh", "ecdsa", "sha256"] }
+#signature = { version = "2.1", optional = true, default-features = false }
```
thibault-martinez: Remove?

src/signatures/secp256k1_ecdsa.rs
```diff
@@ -0,0 +1,210 @@
...
+    }
+}
+
+impl From<Address> for [u8; Address::LENGTH] {
+    fn from(address: Address) -> Self {
+        address.0
+    }
+}
+
+impl Hash for Address {
```
thibault-martinez: Can't this be derived?

src/signatures/secp256k1_ecdsa.rs
```diff
@@ -0,0 +1,210 @@
...
+        k256::ecdsa::SigningKey::from_bytes(&bytes.into())
+            .map_err(|_| crate::Error::ConvertError {
+                from: "bytes",
+                to: "secp256k1 ecdsa secret key",
+            })
+            .map(Self)
+    }
+
+    // Signature operation can fail for some reason.
+    // Not sure what to do. Panic? Retry?
```
thibault-martinez: I don't think we can do anything tbh, just return the error and let the user deal with it. I'd remove the comments.

src/signatures/secp256k1_ecdsa.rs
```diff
@@ -0,0 +1,210 @@
...
+    pub fn try_sign(&self, msg: &[u8]) -> crate::Result<Signature> {
+        self.0
+            .sign_recoverable(msg)
+            .map_err(|_| crate::Error::SignatureError { alg: "secp256k1 ecdsa" })
+            .map(|(sig, rid)| Signature(sig, rid))
+    }
+
+    pub fn sign(&self, msg: &[u8]) -> Signature {
```
thibault-martinez, line 52: Do you understand when/how it can panic? Depending on that, I'm not even sure I would provide this function tbh

src/signatures/secp256k1_ecdsa.rs
```diff
@@ -0,0 +1,210 @@
...
+    type Error = crate::Error;
+    fn try_from(bytes: [u8; PublicKey::LENGTH]) -> crate::Result<Self> {
+        Self::try_from_bytes(&bytes)
+    }
+}
+
+impl Hash for PublicKey {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        let encoded_point = self.0.to_encoded_point(true);
+        encoded_point.as_bytes().hash(state);
```
thibault-martinez, line 137: For my own curiosity, if you could explain the different between encoded point and sec1 bytes please 🙏🏻

src/signatures/secp256k1_ecdsa.rs
```diff
@@ -0,0 +1,210 @@
...
+
+    pub fn to_bytes(self) -> [u8; PublicKey::LENGTH] {
+        let encoded_point = self.0.to_encoded_point(true);
+        let slice = encoded_point.as_bytes();
+        let mut bytes = [0_u8; PublicKey::LENGTH];
+        bytes.copy_from_slice(slice);
+        bytes
+    }
+
+    pub fn try_from_bytes(bytes: &[u8; PublicKey::LENGTH]) -> crate::Result<Self> {
```
thibault-martinez, line 76: `try_from_bytes(self.to_bytes())` would work, right? Because one uses encoded point and the other sec1 bytes. I guess they can be interchangeable but why are we not using the same method for both?

src/signatures/secp256k1_ecdsa.rs
```diff
@@ -0,0 +1,210 @@
+// Copyright 2023 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+use core::convert::TryFrom;
+use core::hash::{Hash, Hasher};
+
+use zeroize::ZeroizeOnDrop;
+
+#[derive(Clone, Copy, Debug, Eq, Ord, PartialEq, PartialOrd)]
+pub struct Address([u8; Self::LENGTH]);
```
thibault-martinez: Should this really be part of crypto? Is it part of the specs? I feel like most of the time it would be down to the consumer to decide how to derive the address from the public key?

src/signatures/secp256k1_ecdsa.rs
```diff
@@ -0,0 +1,210 @@
...
+                to: "Secp256k1 public key",
+            })
+    }
+
+    // credit: [secret_key_to_address](https://github.com/gakonst/ethers-rs/)
+    pub fn to_address(&self) -> Address {
+        // let public_key = secret_key.verifying_key();
+        let public_key = self.0.to_encoded_point(/* compress = */ false);
+        let public_key = public_key.as_bytes();
+        debug_assert_eq!(public_key[0], 0x04);
```
thibault-martinez, line 113: For my curiosity, what does this type byte describe?

Cargo.toml
semenov-vladyslav: Done

src/signatures/secp256k1_ecdsa.rs
```diff
@@ -0,0 +1,210 @@
...
+
+    pub fn to_bytes(&self) -> [u8; Signature::LENGTH] {
+        let mut bytes = [0_u8; Signature::LENGTH];
+        bytes[0..64].copy_from_slice(&self.0.to_bytes());
+        bytes[64] = self.1.into();
+        bytes
+    }
+
+    pub fn try_from_bytes(sig: &[u8; Signature::LENGTH]) -> crate::Result<Self> {
+        let rid = k256::ecdsa::RecoveryId::from_byte(sig[64]).ok_or(crate::Error::ConvertError {
```
thibault-martinez, line 154: `sig[64]` would panic if the bytes don't have the correct size

thibault-martinez: Maybe you can just call `try_from_slice` since it checks the size

Cargo.toml
```diff
@@ -111,6 +113,7 @@ zeroize = { version = "1.5", optional = true, default-features = false, features
 scrypt = { version = "0.11", optional = true, default-features = false }
 hkdf = { version = "0.12", optional = true, default-features = false }
 base64 = { version = "0.21", optional = true, default-features = false }
+k256 = { version = "0.13", optional = true, default-features = false, features = ["ecdh", "ecdsa", "sha256"] }
```
thibault-martinez:
```suggestion
k256 = { version = "0.13", optional = true, default-features = false, features = [ "ecdh", "ecdsa", "sha256" ] }
```

src/signatures/secp256k1_ecdsa.rs
semenov-vladyslav: It sure can

src/signatures/secp256k1_ecdsa.rs
semenov-vladyslav: Removed

src/signatures/secp256k1_ecdsa.rs
semenov-vladyslav: Encoded point is the SEC1 encoded point, it can be in compressed, uncompressed or compact form. I'm not sure about compact form, looks like it loses information about the y coordinate. Compressed SEC1 point seems to do the trick, it is [tag + x-coordinate] where tag=2 encodes even y coord, tag=3 encodes odd y coord. Compressed SEC1 form is used in Ethereum for public keys and in SLIP-10 in non-hardened key derivation.

semenov-vladyslav: I think you meant why I didn't use `to_sec1_bytes()` here and in `PublicKey::to_bytes()`. The reason is to avoid using `Box`, internally `to_sec1_bytes()` it constructs an encoded point with a variable length (because its tag/form can vary). To avoid using `Box` and allocating unnecessary memory I create encoded point directly.

src/signatures/secp256k1_ecdsa.rs
semenov-vladyslav: Good point. Internally, it should be the same. Do you think we should enforce compressed form?

src/signatures/secp256k1_ecdsa.rs
semenov-vladyslav: Where should it be? I understand that it's not directly a part of Secp256k1/ECDSA spec but the whole point was to generate EVM addresses. I haven't found the spec for it, but here's the code that does that https://github.com/gakonst/ethers-rs/blob/f7454f27666750449ca9005a4db032025b4c36ff/ethers-signers/src/aws/utils.rs#L59

src/signatures/secp256k1_ecdsa.rs
semenov-vladyslav: As I mentioned above, SEC1 encoded point has the first byte as tag. Tag=4 encodes uncompressed point (public key).

src/signatures/secp256k1_ecdsa.rs
semenov-vladyslav: `sig[64]` wouldn't panic because `sig: &[u8; Signature::LENGTH]` and `Signature::LENGTH == 65`. Maybe, one method would be enough. I don't like to constantly checking for slice size if its size is known at compile time. But taking fixed-size array from a slice is not very easy.

Cargo.toml
semenov-vladyslav: Done

src/signatures/secp256k1_ecdsa.rs
thibault-martinez: Ah damn sorry, was reading both at the same time and I guess I got confused

src/signatures/secp256k1_ecdsa.rs
semenov-vladyslav: We should enforce it. Otherwise there's two different byte encodings that will result in the same public key: compact form just assumes even y-coord. So encodings [2, even-y-coord] and [5, even-y-coord] correspond to the same public key. This is bad due to malleability reasons.

semenov-vladyslav: Alright. Added checks to only support SEC1 compressed public keys (with tags 2 or 3), compact public keys (tag = 5) are forbidden.

src/signatures/secp256k1_ecdsa.rs
semenov-vladyslav: `Signer` trait has `try_sign` and [`sign`](https://docs.rs/signature/latest/signature/trait.Signer.html#method.sign) methods, the latter can fail. probability of failing for secp256k1 is very small about (1^-127 or so). Types from ecdsa crate implement this trait.

semenov-vladyslav: It can panic if ecdsa produces some bad values during signing. Here's the function that does signing https://github.com/RustCrypto/signatures/blob/master/ecdsa/src/hazmat.rs#L75
It can fail in a couple of places.

tests/secp256k1_ecdsa.rs
```diff
@@ -59,3 +59,18 @@ fn test_secp256k1_sign_verify_random() {
...
+    let mut pk_bytes = [0_u8; 33];
+    sk_bytes[0] += 1;
+    let sk = SecretKey::try_from_bytes(sk_bytes).unwrap();
+    let pk = sk.public_key();
+    pk_bytes.copy_from_slice(&pk.to_bytes());
+    assert_eq!(2, pk_bytes[0]);
+    pk_bytes[0] = 5;
+    // This is a SEC1 Compact form of the same public key. It is forbidden
+    let pk2 = PublicKey::try_from_bytes(&pk_bytes);
+    assert!(pk2.is_err());
```
thibault-martinez, line 75: Actually checking the error would be even better, just to be sure another error is not taking precedence for some reason

src/signatures/secp256k1_ecdsa.rs
thibault-martinez: I mean, I assume how you want to derive an address from a public key could change from one project to another. Which is similar to what we do atm with ED25519, crypto.rs provides us a public key and we hash it with blake2 on our own to derive the address from it. This part is not done in crypto.rs.

tests/secp256k1_ecdsa.rs
semenov-vladyslav: You mean like having separate error values for different error conditions?

thibault-martinez: I mean, you should test that the error is the one you expect, because it could be something else if the test itself is malformed etc...

src/signatures/secp256k1_ecdsa.rs
semenov-vladyslav: Fair enough. So where should it be? In stronghold? If it's outside of crypto then we need to expose uncompressed public key bytes.

src/keys/slip10.rs
```diff
@@ -1,32 +1,87 @@
...
+    Secp256k1,
 }
 
 impl Curve {
+    pub fn is_non_hardened_supported(&self) -> bool {
+        match self {
+            #[cfg(feature = "ed25519")]
+            Curve::Ed25519 => false,
+            #[cfg(feature = "secp256k1")]
+            _ => true,
```
thibault-martinez: I think I would explicitly mention `Curve::Secp256k1`, because if we ever add a new curve type, it will silently return true even though it may be false. Or is there really just `ed25519` that can't support it?

src/keys/slip10.rs
```diff
@@ -1,32 +1,87 @@
...
+        }
+    }
+
     fn seedkey(&self) -> &[u8] {
         match self {
+            #[cfg(feature = "ed25519")]
             Curve::Ed25519 => b"ed25519 seed",
+            #[cfg(feature = "secp256k1")]
+            Curve::Secp256k1 => b"Bitcoin seed",
```
thibault-martinez, line 45: Are these arbitrary from us or is that from some spec?

src/keys/slip10.rs
```diff
@@ -1,32 +1,87 @@
...
+#[derive(ZeroizeOnDrop)]
+pub enum SecretKey {
+    #[cfg(feature = "ed25519")]
+    Ed25519(ed25519::SecretKey),
+    #[cfg(feature = "secp256k1")]
+    Secp256k1Ecdsa(secp256k1_ecdsa::SecretKey),
+}
+
+impl SecretKey {
+    pub fn to_bytes(&self) -> [u8; 32] {
```
thibault-martinez: Will 32 bytes always be true for any potential future curve?

thibault-martinez: Just as a note: this creates a non-zeroize-on-drop copy of the secret key. I wonder if we should just return a slice and let the user do copies if they want to.

src/keys/slip10.rs
```diff
@@ -44,100 +99,404 @@ impl Seed {
...
+            Err(crate::Error::InvalidArgumentError {
+                alg: "SLIP10",
+                expected: "curve supporting non-hardened key derivation",
+            })
+        }
+    }
+}
+
+#[derive(Clone)]
+pub struct ExtendedPublicKey(KeyImpl);
```
thibault-martinez, line 208: Can you elaborate on what is an extended public key? Compared to a public key.

src/keys/slip10.rs
```diff
@@ -44,100 +99,404 @@ impl Seed {
...
+    pub fn public_bytes(&self) -> &[u8; 33] {
+        self.0.public_bytes()
+    }
 
-        Ok(Self(i))
+    pub fn public_key(&self) -> PublicKey {
+        match self.curve() {
+            #[cfg(feature = "ed25519")]
+            Curve::Ed25519 => {
+                unreachable!("SLIP10 does not support non-hardened key derivation from Ed25519 public keys")
```
thibault-martinez, line 246: Unreachable because a `ExtendedPublicKey` can't even be created with Ed25519 ?

src/keys/slip10.rs
```diff
@@ -44,100 +99,404 @@ impl Seed {
...
 
-    pub fn derive(&self, chain: &Chain) -> crate::Result<Key> {
-        let mut k = *self;
-        for c in &chain.0 {
-            k = k.child_key(c)?;
+    pub fn derive(&self, chain: &Chain) -> crate::Result<Self> {
+        #[cfg(feature = "ed25519")]
+        debug_assert_ne!(Curve::Ed25519, self.curve());
+
+        if chain.all_non_hardened() {
```
thibault-martinez, line 277: Nothing at all has to be hardened for Secp256k1?

src/keys/slip10.rs
```diff
@@ -44,100 +99,404 @@ impl Seed {
...
-        if bs.len() != 64 {
-            return Err(crate::Error::BufferSize {
-                name: "key",
-                has: bs.len(),
-                needs: 64,
-            });
+#[derive(Clone)]
+struct KeyImpl {
+    curve: Curve,
+    ext: [u8; 65],
```
thibault-martinez, line 305: What's the extra byte for ?

src/keys/slip10.rs
```diff
@@ -44,100 +99,404 @@ impl Seed {
...
 
-    fn try_from(bs: &[u8]) -> Result<Self, Self::Error> {
-        if bs.len() != 64 {
-            return Err(crate::Error::BufferSize {
-                name: "key",
-                has: bs.len(),
-                needs: 64,
-            });
+#[derive(Clone)]
+struct KeyImpl {
```
thibault-martinez, line 300: So this can be used for both public and private keys?

src/keys/slip10.rs
thibault-martinez, line 305: This is secret/public key + chain code ? Maybe add a small comment about the content of the bytes

thibault-martinez: I saw it later but maybe still explain again pls

src/keys/slip10.rs
```diff
@@ -44,100 +99,404 @@ impl Seed {
...
         }
+    }
 
-        let mut ds = [0; 64];
-        ds.copy_from_slice(bs);
-        Ok(Self(ds))
+    fn is_public_key_valid(&self) -> bool {
+        match self.curve {
+            #[cfg(feature = "ed25519")]
+            Curve::Ed25519 => unreachable!("ed25519 curve is not supported for non-hardened public key derivation"),
```
thibault-martinez, line 367: That seems a bit weird, looks reachable to me?

src/keys/slip10.rs
```diff
@@ -44,100 +99,404 @@ impl Seed {
...
+            Curve::Secp256k1 => {
+                (self.ext[0] == 2 || self.ext[0] == 3) && k256::PublicKey::from_sec1_bytes(self.public_bytes()).is_ok()
+            }
+        }
+    }
+
+    fn calc_public_bytes(&self) -> [u8; 33] {
+        match self.curve {
+            #[cfg(feature = "ed25519")]
+            Curve::Ed25519 => unreachable!(),
```
thibault-martinez, line 378: same?

src/keys/slip10.rs
```diff
@@ -44,100 +99,404 @@ impl Seed {
...
+        }
+
+        data.zeroize();
+        key
     }
 }
 
 #[derive(Clone, Copy, Debug, Eq, Hash, PartialEq, Serialize, Deserialize)]
-pub struct Segment {
-    hardened: bool,
```
thibault-martinez, line 123: the bool was always useless because it can be deduced from the bs?

src/keys/slip10.rs
Thoralf-M: It's here https://github.com/satoshilabs/slips/blob/master/slip-0010.md#master-key-generation

thibault-martinez: Ah thanks

tests/secp256k1_ecdsa.rs
semenov-vladyslav: The errors are not very meaningful, `try_from_bytes` returns `ConvertError`. Probably need to test for correct errors in other tests too.

src/keys/slip10.rs
semenov-vladyslav: SLIP-10 applies to 3 curves (ed25519, secp256k1, nist p-256). Only ed25519 doesn't support non-hardened key derivation. In general, it's best to explicitly have patterns for Curve constructors, otherwise `#[cfg(...)]` condition would be quite complex and ugly.

semenov-vladyslav: Done

src/keys/slip10.rs
semenov-vladyslav: They are defined [in the slip](https://github.com/satoshilabs/slips/blob/master/slip-0010.md?plain=1#L70)

semenov-vladyslav: They are defined [in the slip](https://github.com/satoshilabs/slips/blob/master/slip-0010.md?plain=1#L70)

src/keys/slip10.rs
semenov-vladyslav: So far, slip-10 supports only 256-bit curves, so this looks pretty stable. Otherwise we'd return some `Zeroizing<Vec<u8>>`.

semenov-vladyslav: `secp256k1_ecdsa::SecretKey` does not support `as_ref()` method because internal implementation doesn't support it. Not sure if we even need to have this method in the first place. Users should pattern match on value to access the actual secret key value.

src/keys/slip10.rs
semenov-vladyslav: Extended public key contains public key and chain code (and a curve). It allows it to derive child non-hardened public keys. It is the terminology [from the slip](https://github.com/satoshilabs/slips/blob/master/slip-0010.md?plain=1#L115).

src/keys/slip10.rs
semenov-vladyslav: Yes. `ExtendedPublicKey` constructors are guarded by `curve.is_non_hardened_supported()` which fails for Ed25519.

src/keys/slip10.rs
semenov-vladyslav: Child public key can be derived from extended master public key only with non-hardened segment. Hardened segments require knowledge of the master private key which is obviously not available.

src/keys/slip10.rs
semenov-vladyslav: KeyImpl is polymorphic in two ways: for different curves and for secret/public key derivation. For secret key derivation the first byte is zero, the next 32 bytes are secret key, the last 32 bytes are chain code. For public key derivation the first 33 bytes are SEC1 compressed encoding of the public key point (the first byte is tag value 2 or 3 depending whether y coordinate is even or odd, the next 32 bytes is x coordinate), the last 32 bytes is again chain code.

src/keys/slip10.rs
semenov-vladyslav: Yes. Otherwise I'd need to copy-paste pieces of code or use some traits which would make code more complex.

src/keys/slip10.rs
semenov-vladyslav: Done.

src/keys/slip10.rs
semenov-vladyslav: Well, it shouldn't. It is only used in `ExtendedPublicKey::try_from_extended_bytes` but after the `curve.is_non_hardened_supported()` guard. So it should be safe.

src/keys/slip10.rs
semenov-vladyslav: Sort of. All paths leading to this should be guarded by `is_non_hardened_supported()`.

src/keys/slip10.rs
semenov-vladyslav: Yes. And there's no need to convert u32 to [u8; 4] early.

src/keys/slip10.rs
thibault-martinez: If we keep it, I would at least leave a comment saying it should be cleared

semenov-vladyslav: I'd prefer using Zeroizing, that'll be safer by default.

src/signatures/secp256k1_ecdsa.rs
thibault-martinez: Stronghold itself doesn't really need an address since it only delivers the public key to the consumer. For ED25519 we derive the address in the SDK directly so I guess we should do the same? https://github.com/iotaledger/iota-sdk/blob/develop/sdk/src/client/stronghold/secret.rs#L87

tests/secp256k1_ecdsa.rs
thibault-martinez: Yeah I guess it's contextual. Maybe it doesn't make sense for this one idk. Just saying that if we don't check the error, it could actually be another one due to an unrelated reason. But no worries.

thibault-martinez (approved):

Thoralf-M (approved):

PR #188, opened 2023-05-03
Allow `no_std` serde

DaughterOfMars:
# Description of change

The `serde` dependency used default features, which includes `std`, thus breaking no_std. 

Additionally, the `[dev-dependencies]` also cause the build to fail on `no_std` targets. A new resolver is added to resolve this.

## Type of change

- Bug fix (a non-breaking change which fixes an issue)

Build using

```sh
cargo build  --target riscv64gc-unknown-none-elf
```

with various feature combinations. Note that the `random` feature will not build on some targets, including the above.

### DIFF
--- a/.changes/serde-no-std
+++ b/.changes/serde-no-std
@@ -1,0 +1,5 @@
+---
+"iota-crypto": minor
+---
+
+Make `serde` dependency `no_std` compatible.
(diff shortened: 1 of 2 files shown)


.changes/serde-no-std
```diff
@@ -0,0 +1,5 @@
+---
+"iota-crypto": minor
```
thibault-martinez, line 2: patch?

thibault-martinez: Meh, maybe the resolver is doing more that we can see, forget about it

thibault-martinez (approved):

PR #193, opened 2023-05-17
Fix `no_std` serde

DaughterOfMars:
# Description of change

The serde dependency cannot use alloc types without the feature, so this adds that feature. I also updated the rust edition while I was changing the cargo file.

## Type of change


thibault-martinez (commented):
Needs a .change for a release

semenov-vladyslav (approved):

thibault-martinez (approved):

Issue #192, opened 2023-05-10
Make u32 from Segment pub reachable


Thoralf-M:
## Description

https://github.com/iotaledger/crypto.rs/blob/730b64599eb72a04422b1ee81d68413715881d23/src/keys/slip10.rs#L490

## Motivation

In iota-sdk we want to get it without a useless roundtrip from .bs()

## Requirements

1. Add a getter/deref so the number can be read

## Are you planning to do it yourself in a pull request?

No.

PR #194, opened 2023-05-17
Slip10 rewrite

semenov-vladyslav:
# Description of change

Please write a summary of your changes and why you made them.

## Type of change

- [ ] Documentation Fix

## Change checklist

### DIFF
--- a/.changes/slip10-rewrite.md
+++ b/.changes/slip10-rewrite.md
@@ -1,5 +1,5 @@
 ---
-"iota-crypto": patch
+"iota-crypto": minor
 ---
 
 Rewrite SLIP-10 in a more generic form.
--- a/.changes/slip10-rewrite.md
+++ b/.changes/slip10-rewrite.md
@@ -1,0 +1,5 @@
+---
+"iota-crypto": minor
+---
+
+Rewrite SLIP-10 in a more generic form.
--- a/.changes/slip10-rewrite.md
+++ b/.changes/slip10-rewrite.md
@@ -1,5 +1,5 @@
 ---
-"iota-crypto": minor
+"iota-crypto": patch
 ---
 
 Rewrite SLIP-10 in a more generic form.
--- a/src/keys/slip10.rs
+++ b/src/keys/slip10.rs
@@ -55,7 +55,7 @@ pub mod ed25519 {
             true
         }
         fn calc_non_hardened_data(_key_bytes: &[u8; 33]) -> [u8; 33] {
-            unreachable!("SLIP10: ed25519 does no support non-hardened key derivation")
+            unreachable!("SLIP10: ed25519 does not support non-hardened key derivation")
         }
     }
 
--- a/src/keys/slip10.rs
+++ b/src/keys/slip10.rs
@@ -28,7 +28,7 @@ pub mod hazmat {
 
     pub trait Derivable {
         type Curve: Curve;
-        const IS_NON_HARDENED_SUPPORTED: bool;
+        const ALLOW_NON_HARDENED: bool;
         type Key;
         fn is_key_valid(key_bytes: &[u8; 33]) -> bool;
         fn into_key(key_bytes: &[u8; 33]) -> Self::Key;
@@ -53,7 +53,7 @@ pub mod ed25519 {
 
     impl Derivable for SecretKey<Ed25519> {
         type Curve = Ed25519;
-        const IS_NON_HARDENED_SUPPORTED: bool = false;
+        const ALLOW_NON_HARDENED: bool = false;
         type Key = ed25519::SecretKey;
         fn is_key_valid(key_bytes: &[u8; 33]) -> bool {
             key_bytes[0] == 0
@@ -91,7 +91,7 @@ pub mod secp256k1 {
 
     impl Derivable for SecretKey<Secp256k1> {
         type Curve = Secp256k1;
-        const IS_NON_HARDENED_SUPPORTED: bool = true;
+        const ALLOW_NON_HARDENED: bool = true;
         type Key = secp256k1_ecdsa::SecretKey;
         fn is_key_valid(key_bytes: &[u8; 33]) -> bool {
             debug_assert_eq!(0, key_bytes[0]);
@@ -142,7 +142,7 @@ pub mod secp256k1 {
 
     impl Derivable for PublicKey<Secp256k1> {
         type Curve = Secp256k1;
-        const IS_NON_HARDENED_SUPPORTED: bool = true;
+        const ALLOW_NON_HARDENED: bool = true;
         type Key = secp256k1_ecdsa::PublicKey;
         fn is_key_valid(key_bytes: &[u8; 33]) -> bool {
             (key_bytes[0] == 2 || key_bytes[0] == 3) && k256::PublicKey::from_sec1_bytes(key_bytes).is_ok()
@@ -265,7 +265,7 @@ where
             ext: [0; 65],
         };
         HMAC_SHA512(&seed.0, C::SEEDKEY, key.ext_mut());
-        while !key.is_secret_key_valid() {
+        while !key.is_key_valid() {
             let mut tmp = [0_u8; 64];
             tmp.copy_from_slice(&key.ext[1..]);
             HMAC_SHA512(&tmp, C::SEEDKEY, key.ext_mut());
@@ -364,7 +364,7 @@ impl<K: hazmat::Derivable> Extended<K> {
     }
 
     pub fn derive(&self, chain: &Chain) -> crate::Result<Self> {
-        if K::IS_NON_HARDENED_SUPPORTED || chain.all_hardened() {
+        if K::ALLOW_NON_HARDENED || chain.all_hardened() {
             let mut key: Self = self.clone();
             for segment in &chain.0 {
                 key = key.derive_child_key(segment);
@@ -373,18 +373,18 @@ impl<K: hazmat::Derivable> Extended<K> {
         } else {
             Err(crate::Error::InvalidArgumentError {
                 alg: "SLIP10",
-                expected: "hardened key index for Ed25519 master secret key",
+                expected: "hardened key index",
             })
         }
     }
 
     pub fn child_key(&self, segment: &Segment) -> crate::Result<Self> {
-        if K::IS_NON_HARDENED_SUPPORTED || segment.is_hardened() {
+        if K::ALLOW_NON_HARDENED || segment.is_hardened() {
             Ok(self.derive_child_key(segment))
         } else {
             Err(crate::Error::InvalidArgumentError {
                 alg: "SLIP10",
-                expected: "hardened key index for Ed25519 master secret key",
+                expected: "hardened key index",
             })
         }
     }
@@ -401,7 +401,7 @@ impl<K: hazmat::Derivable> Extended<K> {
         unsafe { &mut *(self.ext[..33].as_mut_ptr() as *mut [u8; 33]) }
     }
 
-    fn is_secret_key_valid(&self) -> bool {
+    fn is_key_valid(&self) -> bool {
         K::is_key_valid(self.key_bytes())
     }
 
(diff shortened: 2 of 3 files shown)


src/keys/slip10.rs
```diff
@@ -485,6 +394,13 @@ impl KeyImpl {
...
 
+impl<K: hazmat::Derivable> TryFrom<&[u8; 65]> for Extended<K> {
+    type Error = crate::Error;
+    fn try_from(ext_bytes: &[u8; 65]) -> crate::Result<Self> {
+        Self::try_from_extended_bytes(ext_bytes)
+    }
+}
+
 #[derive(Clone, Copy, Debug, Eq, Hash, PartialEq, Serialize, Deserialize)]
 pub struct Segment(u32);
```
Thoralf-M: Can you add a getter/deref for the u32 here?
https://github.com/iotaledger/crypto.rs/issues/192

.changes/slip10-rewrite.md
```diff
@@ -0,0 +1,5 @@
+---
+"iota-crypto": minor
```
thibault-martinez, line 2: Should be a patch?

DaughterOfMars (approved):

src/keys/slip10.rs
semenov-vladyslav: What would be preferable solution?
1. Get rid of Segment type and use u32 directly.
2. Make inner field public.
3. Add accessor `fn into_u32(&self) -> u32;`.
4. `impl Deref/DerefMut for Segment`.
5. `impl From<u32> for Segment`, `impl From<Segment> for u32`.

.changes/slip10-rewrite.md
semenov-vladyslav: Slip10 API is breaking compared to 0.19.

src/keys/slip10.rs
DaughterOfMars: If there's no specific reason to define the type, then remove it. Otherwise, I would just make the inner field pub.

src/keys/slip10.rs
```diff
@@ -463,7 +475,7 @@ impl Chain {
     }
 
     pub fn all_non_hardened(&self) -> bool {
-        self.0.iter().all(|s| !s.is_hardened())
+        self.0.iter().all(Segment::is_non_hardened)
```
DaughterOfMars: More idiomatic to do
```
!self.0.iter().any(Segment::is_hardened)
```

src/keys/slip10.rs
semenov-vladyslav: Made field pub, added From impl.

.changes/slip10-rewrite.md
semenov-vladyslav: Patch it is.

src/keys/slip10.rs
semenov-vladyslav: Thanks, already added `Segment::is_non_hardened`.

src/keys/slip10.rs
```diff
@@ -6,80 +6,170 @@
 use alloc::vec::Vec;
 use core::convert::TryFrom;
 
+// TODO: derive(Serialize, Deserialize) for Chain
```
Thoralf-M: That's already done?
```suggestion
```

src/keys/slip10.rs
semenov-vladyslav: Done.

.changes/slip10-rewrite.md
semenov-vladyslav: It's not. It's minor after all.

src/keys/slip10.rs
semenov-vladyslav: Done, removed comment.

src/keys/slip10.rs
```diff
@@ -485,25 +393,44 @@ impl KeyImpl {
...
 impl Segment {
-    pub fn from_u32(i: u32) -> Self {
-        Self(i)
-    }
-
     pub fn is_hardened(&self) -> bool {
         self.0 & Self::HARDEN_MASK != 0
     }
 
+    pub fn is_non_hardened(&self) -> bool {
```
DaughterOfMars: Can we remove this now?

src/keys/slip10.rs
```diff
@@ -7,79 +7,168 @@ use alloc::vec::Vec;
...
 // https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki
 // https://en.bitcoin.it/wiki/BIP_0039
 
-#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq, Serialize, Deserialize)]
-pub enum Curve {
-    #[cfg(feature = "ed25519")]
-    Ed25519,
-    #[cfg(feature = "secp256k1")]
-    Secp256k1,
+pub mod hazmat {
```
thibault-martinez: Just curious, are these traits expected to be used externally?

DaughterOfMars: Good point. No reason for the internal mod.

semenov-vladyslav: No, but they end up in public API so there's no way to make them private.

src/keys/slip10.rs
semenov-vladyslav: Done.

src/keys/slip10.rs
```diff
@@ -7,79 +7,168 @@ use alloc::vec::Vec;
...
+    pub trait Derivable {
+        fn is_key_valid(key_bytes: &[u8; 33]) -> bool;
+        fn into_key(key_bytes: &[u8; 33]) -> Self;
```
thibault-martinez: I'm nitpicking a bit but Rust usual naming convention for `into_` is that it consumes the parameter, otherwise it's a `to_`

src/keys/slip10.rs
```diff
@@ -7,79 +7,168 @@ use alloc::vec::Vec;
...
+        fn into_key(key_bytes: &[u8; 33]) -> Self {
+            debug_assert_eq!(0, key_bytes[0]);
+            let sk_bytes: &[u8; 32] = unsafe { &*(key_bytes[1..].as_ptr() as *const [u8; 32]) };
+            ed25519::SecretKey::from_bytes(sk_bytes)
+        }
+        fn add_key(_key_bytes: &mut [u8; 33], _parent_key: &[u8; 33]) -> bool {
+            true
+        }
+        fn calc_non_hardened_data(_key_bytes: &[u8; 33]) -> [u8; 33] {
+            unreachable!()
```
thibault-martinez: If both the trait and the type are public, this actually is reachable? Maybe at least put a message in the panic?

DaughterOfMars: Maybe instead there should be a separate trait, not implemented for this type.

src/keys/slip10.rs
semenov-vladyslav: Noted.

src/keys/slip10.rs
semenov-vladyslav: Done.

src/keys/slip10.rs
thibault-martinez: FYI https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv

src/keys/slip10.rs
semenov-vladyslav: I don't know how to abstract it away with a trait. Need to think about it a bit more.

src/keys/slip10.rs
```diff
@@ -7,79 +7,168 @@ use alloc::vec::Vec;
...
+        fn to_key(key_bytes: &[u8; 33]) -> Self {
+            debug_assert_eq!(0, key_bytes[0]);
+            let sk_bytes: &[u8; 32] = unsafe { &*(key_bytes[1..].as_ptr() as *const [u8; 32]) };
+            ed25519::SecretKey::from_bytes(sk_bytes)
+        }
+        fn add_key(_key_bytes: &mut [u8; 33], _parent_key: &[u8; 33]) -> bool {
+            true
+        }
+        fn calc_non_hardened_data(_key_bytes: &[u8; 33]) -> [u8; 33] {
+            unreachable!("SLIP10: ed25519 does no support non-hardened key derivation")
```
thibault-martinez:
```suggestion
            panic!("SLIP10: ed25519 does not support non-hardened key derivation")
```

src/keys/slip10.rs
```diff
@@ -99,384 +188,203 @@ impl Seed {
...
-        self.0.curve
+impl<K> Zeroize for Extended<K> {
+    fn zeroize(&mut self) {
+        self.ext.zeroize()
     }
+}
 
-    pub fn extended_bytes(&self) -> &[u8; 64] {
-        self.0.extended_secret_bytes()
+impl<K> Drop for Extended<K> {
```
thibault-martinez: Wouldn't that be done by `ZeroizeOnDrop` ?

semenov-vladyslav: Good point. Let me check.

src/keys/slip10.rs
semenov-vladyslav: Fixed that.

src/keys/slip10.rs
DaughterOfMars: You would create a new trait with that fn and impl it only for types that can perform that fn.

semenov-vladyslav: I should be able to call or not call that method depending on runtime value (segment.is_hardened()). But trait resolution is done at compile time. Maybe, derive method should be generic in some way.

src/keys/slip10.rs
```diff
@@ -99,384 +188,196 @@ impl Seed {
...
 
-#[derive(Clone, Zeroize, ZeroizeOnDrop)]
-pub struct ExtendedSecretKey(KeyImpl);
-
-impl ExtendedSecretKey {
-    pub fn from_seed(curve: Curve, seed: &Seed) -> Self {
-        Self(KeyImpl::from_seed(curve, &seed.0))
-    }
+#[derive(ZeroizeOnDrop)]
+pub struct Extended<K> {
```
thibault-martinez: Is `Extended` a descriptive enough name?

semenov-vladyslav: I thought so, SLIP10 uses terms "extended key", "extended private key", "extended public key".

thibault-martinez: Can we do ExtendedKey then? Just looking at `Extended`, I'm not sure what it refers to

src/keys/slip10.rs
```diff
@@ -99,384 +188,196 @@ impl Seed {
...
-                expected: "hardened key index for Ed25519 master secret key",
-            });
-        }
-
-        let mut key = self.0.clone();
-        for segment in &chain.0 {
-            key = key.child_key(segment);
-        }
-        Ok(Self(key))
+    pub fn into_extended_public_key(&self) -> Extended<K::PublicKey>
```
thibault-martinez: Not an into_

semenov-vladyslav: Right

src/keys/slip10.rs
```diff
@@ -99,384 +188,196 @@ impl Seed {
...
-            } else {
-                Ok(Self(key))
-            }
-        }
+impl<K: hazmat::IsPublicKey> Extended<K> {
+    pub fn from_extended_secret_key(esk: &Extended<K::SecretKey>) -> Self {
+        let mut k = Self {
+            key: core::marker::PhantomData,
+            ext: [0_u8; 65],
+        };
```
thibault-martinez: I see you doing this at different places, can it not be a default impl?

semenov-vladyslav: It can't be Default - it wouldn't represent a valid inner key. I can factor it out into a private `new` method.

src/keys/slip10.rs
DaughterOfMars: But the whole point of the changes was to give a concrete type, so there is no runtime checking.

src/keys/slip10.rs
semenov-vladyslav: The idea was it'll be used with concrete types, eg. `Extended<ed25519::SecretKey>`, `Extended<secp256k1_ecdsa::PublicKey>`. No need to duplicate "Key" twice. I can still change to `ExtendedKey` if you insist.

src/keys/slip10.rs
semenov-vladyslav: Added `new` method.

src/keys/slip10.rs
semenov-vladyslav: Renamed to `Slip10`.

DaughterOfMars (approved):

thibault-martinez (approved):

PR #196, opened 2023-05-23
Additional changes for slip10 rewrite

DaughterOfMars:
I spent some time thinking of ways to make the process of derivation fully statically typed to avoid errant panics. The result isn't perfect, but I think the compromise is probably worth it.


DaughterOfMars:
Closing for now, as the problems are difficult to solve cleanly.

PR #197, opened 2023-05-25
fix: use separator from `WordList` when decoding

qrayven:
# Description of change

the BIP39 implementation uses `WordList` as a source for mnemonics: https://github.com/iotaledger/crypto.rs/blob/9d155ec7b2d8a239d2bf4dae658aadc83e1b69e4/src/keys/bip39.rs#L33 `WordList` also defines the word separator. When `encode()` the given separator is used to generate the mnemonic phrase. When `decode()` the `split_whitespace()` is used which makes the decoding fail for non-whitespace separators.

Additionally, `split_whitespace()` allows repetition of the separator (`\s+`), which causes two side-effects:
 -  when double space is used `verify()` passes, but produced seed is different from the phrase with one space
 -  edge case: when the word starts with a separator character, the `decode()` would fail 

## Links to any relevant issues

This  PR also closes  #125  and fixes https://github.com/iotaledger/iota-sdk/issues/421

## Type of change

Choose a type of change, and delete any options that are not relevant.

- [x] Bug fix (a non-breaking change which fixes an issue)
- [ ] Enhancement (a non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation Fix

## Change checklist

Add an `x` to the boxes that are relevant to your changes, and delete any items that are not.

- [x] I have followed the contribution guidelines for this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

PR #200, opened 2023-05-26
Bip39 safety wrappers

semenov-vladyslav:
# Description of change

The current BIP39 implementation is not ideal in a number of ways as discussed in [here](https://github.com/iotaledger/crypto.rs/pull/197#issuecomment-1562917117). This PR aims to fix those issues: make impl more type safe (to encapsulate normalization checks) and more secrecy aware (avoid allocating unnecessary memory for temporary secret values which might leak and zeroize those that are actually needed).

## Type of change

- [ ] Documentation Fix

## Change checklist


src/keys/bip39.rs
```diff
@@ -5,24 +5,132 @@
...
+    }
+}
+
+impl<'a> PartialEq<str> for MnemonicRef<'a> {
+    fn eq(&self, other: &str) -> bool {
+        self.0.eq(other)
+    }
+}
+
+#[derive(ZeroizeOnDrop)]
```
Thoralf-M: Can we add Clone here?

src/keys/bip39.rs
```diff
@@ -5,24 +5,132 @@
...
+    fn try_from(passphrase_str: &'a str) -> Result<Self, Error> {
+        if is_nfkd(passphrase_str) {
+            Ok(PassphraseRef(passphrase_str))
+        } else {
+            Err(Error::UnnormalizedPassphrase)
+        }
+    }
+}
 
+#[derive(ZeroizeOnDrop)]
```
Thoralf-M: Can we add Clone here?

tests/bip39.rs
```diff
@@ -17,33 +16,39 @@ struct TestVector {
     seed: &'static str,
 }
 
+#[test]
+fn test_separator() {
+    use unicode_normalization::UnicodeNormalization;
+    let s = "　";
+    // assert!(is_nfkd(s));
```
Thoralf-M: What's up with this?

tests/bip39.rs
```diff
@@ -61,54 +66,56 @@ fn test_wordlist_codec() {
...
+//             mnemonic_phrase
+//         );
+//         assert_eq!(
+//             wordlist::verify(mnemonic, &wordlist::JAPANESE).is_ok(),
+//             case.1,
+//             "{}",
+//             mnemonic_phrase
+//         );
+//     }
+// }
```
Thoralf-M: Why commented?

src/keys/bip39.rs
```diff
@@ -43,17 +151,39 @@ pub mod wordlist {
...
+        }
+
+        pub fn new(separator: char, words: &'a [&'a str; 2048]) -> Result<Self, Error> {
+            // normalize separator char
+            let mut s = String::new();
+            s.push(separator);
+            let s: String = s.nfkd().collect();
+
+            if s.chars().count() == 1 {
+                let separator = s.chars().next().unwrap();
```
thibault-martinez: Do an `if let Some(char) = s.chars().next()` instead if .count() == 1 then unwrap

src/keys/bip39.rs
semenov-vladyslav: I guess so. But mnemonic is a secret and we discourage cloning secrets. What would be the use-case for cloning mnemonics?

src/keys/bip39.rs
semenov-vladyslav: Same as for mnemonic. Why do we need to clone passwords?

Thoralf-M: We sometimes do this, for example here when it's needed in different functions https://github.com/iotaledger/iota-sdk/blob/d1ae85026129ce6c3f439d539dc69a4794c751b3/cli/src/command/wallet.rs#L109-L111

tests/bip39.rs
semenov-vladyslav: Fixed that. `s` contains space in non-normalized form.

src/keys/bip39.rs
Thoralf-M: Currently we do this in a test where one function consumes it and we want to check the same operation again (we could also create a new Mnemonic(), but is that better if we end up cloning the String we put in the Mnemonic?)

tests/bip39.rs
semenov-vladyslav: This test became irrelevant. Forgot to remove it at once. Removed.

src/keys/bip39.rs
semenov-vladyslav: This is a workaround to normalize a single char. unicode_normalization crate doesn't provide such function. Here we want to match "exactly one normalized char", not "at least one char". Rewritten

src/keys/bip39.rs
semenov-vladyslav: I'd say it's ok for tests to clone strings. But added derive Clone anyway.

src/keys/bip39.rs
semenov-vladyslav: I'm not sure about `let password = password.into();`. This just makes another copy of password. Why do you need `impl Into<Password>`? Are there many types convertible into `Password`? Why not just pass passwords by reference? Passwords should not be stored internally (to justify moving value), they should be used to compute keys and encrypt/decrypt (to do that you only need a reference). On the other hand, if a password should be stored then it should be stored in stronghold. The best type for passwords if they are passed by reference is `&[u8]`. In bip39 passwords need to be normalized, hence a separate wrapper.

Thoralf-M: Sorry my example was about another password type we have, not the bip39 one

semenov-vladyslav:
Also, there should probably be mnemonic sanitization (as implemented in https://github.com/iotaledger/iota-sdk/pull/471) -- trim leading/trailing/double spaces, check words to be included in wordlist. Otherwise we accept incorrect mnemonics that do not correspond to entropy.

semenov-vladyslav:
Added [`#[non_exhaustive]`](https://doc.rust-lang.org/reference/attributes/type_system.html#the-non_exhaustive-attribute) attribute to `crypto::Error` enum. This is due to some enum patterns being conditionally enabled depending on presence of some features. This might make pattern matching on error values non-exhaustive outside of defining crate. It is needed to keep compatibility if one crate uses (indirectly) crypto with different features. There are limitations on usage of `non_exhaustive` enums -- they can't be constructed outside but this is Ok since we only produce errors in crypto. Also, when pattern matching "match any" pattern (`_ => ...`) should be used.

This fix deserves its own PR probably, but it's just one line and it really fixed the potential issue with errors.

src/keys/bip39.rs
```diff
@@ -43,17 +165,43 @@ pub mod wordlist {
...
+                    words.iter().try_for_each(|word| {
+                        if is_nfkd(word) && !word.contains(separator) {
+                            Ok(())
+                        } else {
+                            Err(Error::BadWordlist)
+                        }
+                    })?;
+                    Ok(Self { words, separator })
+                } else {
+                    Err(Error::BadSeparator)
```
qrayven, line 262: This erros says, that separator is valid only when it can be normalized to one character.
But for instance `Å` is being normalized to two charcters and wont be valid in this case. Is this expected behavior?

semenov-vladyslav: Yes. I think one-char separator covers most real-world use-cases (eg. space, tab, new line, even this weird japanese space normalizes to the ASCII space). If we need to support multi-char separator then this definitely changes matters.

tests/bip39.rs
```diff
@@ -61,54 +66,24 @@ fn test_wordlist_codec() {
...
+        assert_eq!(wordlist::verify((&ms).into(), ws), Ok(()));
     }
 }
 
 #[test]
 fn test_mnemonic_phrase_additional_whitespace() {
     // additional whitespace at the beginning
     assert!(
-        wordlist::decode(" sand luggage rack used middle crater deal scare high ring swim fish use then video visa can foot clog base quality all elephant retreat", &wordlist::ENGLISH).is_err(),
+        wordlist::decode(" sand luggage rack used middle crater deal scare high ring swim fish use then video visa can foot clog base quality all elephant retreat".try_into().unwrap(), &wordlist::ENGLISH).is_err(),
```
thibault-martinez: I know it wasn't there, but can you test that the error is the expected one?

src/keys/bip39.rs
```diff
@@ -43,17 +165,43 @@ pub mod wordlist {
...
-        NoSuchWord(String),
-        ChecksumMismatch,
-    }
-
     const fn cs(ent: usize) -> usize {
         ent / 32
     }
 
+    impl<'a> Wordlist<'a> {
+        // TODO: should it be pub?
```
thibault-martinez: I'd say no, so I'd remove the comment

tests/bip39.rs
semenov-vladyslav: Good point. I didn't touch this subject in the PR. We should probably test for different error cases.

src/keys/bip39.rs
```diff
@@ -43,17 +165,43 @@ pub mod wordlist {
...
+        const fn new_unchecked(separator: char, words: &'a [&'a str; 2048]) -> Self {
+            Self { words, separator }
+        }
+
+        pub fn new(separator: char, words: &'a [&'a str; 2048]) -> Result<Self, Error> {
```
thibault-martinez, line 237: Just curious, what's the usecase of actually taking a separator as parameter? Can't we just use one internally?

src/keys/bip39.jp.rs
```diff
@@ -3,9 +3,9 @@
 
 // https://raw.githubusercontent.com/bitcoin/bips/master/bip-0039/japanese.txt
 
-pub const JAPANESE: Wordlist = Wordlist {
-    separator: "　",
-    words: &[
+pub const JAPANESE: Wordlist = Wordlist::new_unchecked(
+    ' ', //'　', // U+3000 is normalized to U+20
```
thibault-martinez, line 7: Why a comment here and not in english?

src/keys/bip39.rs
qrayven: I'm ok with it. I would make additional comments to the description tough. For future reference - if someone would like to use it in a super weird way ;)

src/keys/bip39.en.rs
```diff
@@ -3,9 +3,9 @@
 
 // https://raw.githubusercontent.com/bitcoin/bips/master/bip-0039/english.txt
 
-pub const ENGLISH: Wordlist = Wordlist {
-    separator: " ",
-    words: &[
+pub const ENGLISH: Wordlist = Wordlist::new_unchecked(
```
kwek20, line 6: why not use the rust crate? https://github.com/rust-bitcoin/rust-bip39/blob/master/src/language/english.rs

src/keys/bip39.rs
```diff
@@ -5,24 +5,146 @@
...
+pub struct Seed([u8; 64]);
+
+impl AsRef<[u8; 64]> for Seed {
+    fn as_ref(&self) -> &[u8; 64] {
+        &self.0
+    }
+}
+
+impl Default for Seed {
+    fn default() -> Self {
```
kwek20: i think a default seed sounds off, lets rename it to null or something

src/keys/bip39.rs
```diff
@@ -5,24 +5,146 @@
...
+}
+
+impl<'a> PartialEq<str> for MnemonicRef<'a> {
+    fn eq(&self, other: &str) -> bool {
+        self.0.eq(other)
+    }
+}
+
+#[derive(Clone, ZeroizeOnDrop)]
+pub struct Mnemonic(String);
```
Alex6323: should we also derive `Eq` , `PartialEq`, and `Zeroize`?

src/keys/bip39.rs
```diff
@@ -5,24 +5,146 @@
...
+
+impl Deref for Mnemonic {
+    type Target = str;
+    fn deref(&self) -> &str {
+        &self.0
+    }
+}
+
+impl From<&str> for Mnemonic {
```
Alex6323: I feel like this could cause some security issues with users creating the mnemonic from a `&str` while possessing ownership of a `String` which they don't use any further but forgetting to zeroize that original. So I probably would change this to a `From<String>` and claim the ownership. The user would be forced to clone the original and make a conscious decision about it (people are mostly averse to cloning and don't do that lightly, also clippy will warn them about unnecesary clones)

src/keys/bip39.rs
```diff
@@ -5,24 +5,146 @@
...
+impl<'a> PartialEq<str> for MnemonicRef<'a> {
+    fn eq(&self, other: &str) -> bool {
+        self.0.eq(other)
+    }
+}
+
+#[derive(Clone, ZeroizeOnDrop)]
+pub struct Mnemonic(String);
+
+impl Deref for Mnemonic {
```
Alex6323: ~Here I probably would impl `AsRef<str>` instead.~ Is this still necessary when we have `AsRef<str>` already?

src/keys/bip39.rs
```diff
@@ -5,24 +5,146 @@
...
+        if is_nfkd(passphrase_str) {
+            Ok(PassphraseRef(passphrase_str))
+        } else {
+            Err(Error::UnnormalizedPassphrase)
+        }
+    }
+}
+
+#[derive(Clone, ZeroizeOnDrop)]
+pub struct Passphrase(String);
```
Alex6323: also derive ~`Eq`, `PartialEq`,~ `Zeroize`?

src/keys/bip39.rs
```diff
@@ -5,24 +5,146 @@
...
+
+#[derive(Clone, Copy, PartialEq, Eq)]
+pub struct PassphraseRef<'a>(&'a str);
+
+impl<'a> Deref for PassphraseRef<'a> {
+    type Target = str;
+    fn deref(&self) -> &str {
+        self.0
+    }
+}
```
Alex6323, line 141: `AsRef<str>` instead?

src/keys/bip39.rs
```diff
@@ -5,24 +5,146 @@
...
+use core::ops::Deref;
 
-use unicode_normalization::UnicodeNormalization;
+use unicode_normalization::{is_nfkd, UnicodeNormalization};
+use zeroize::{Zeroize, ZeroizeOnDrop, Zeroizing};
 
-pub fn mnemonic_to_seed(m: &Mnemonic, p: &Passphrase, s: &mut Seed) {
-    let m = m.chars().nfkd().collect::<String>();
+#[derive(Debug, PartialEq, Eq)]
+pub enum Error {
```
Alex6323, line 19: docs? (also for the other types in this module)

src/keys/bip39.rs
```diff
@@ -5,24 +5,146 @@
...
+pub struct Passphrase(String);
+
+impl Deref for Passphrase {
+    type Target = str;
+    fn deref(&self) -> &str {
+        &self.0
+    }
+}
+
+impl From<&str> for Passphrase {
```
Alex6323: same here what I mentioned for the `Mnemonic`. `From<String>` will probably be no less convenient but would prevent user from accidentally leaking the original.

src/keys/bip39.rs
```diff
@@ -5,24 +5,146 @@
...
+    }
+}
+
+impl<'a> From<&'a Passphrase> for PassphraseRef<'a> {
+    fn from(passphrase_ref: &'a Passphrase) -> Self {
+        Self(&passphrase_ref.0)
+    }
+}
+
+#[derive(Clone, ZeroizeOnDrop)]
```
Alex6323: can probably also derive `Eq`, `PartialEq`, and `Zeroize`.

src/keys/bip39.rs
```diff
@@ -5,24 +5,146 @@
...
+        Self([0_u8; 64])
+    }
+}
+
+pub fn mnemonic_to_seed(m: MnemonicRef, p: PassphraseRef, s: &mut Seed) {
     let mut salt = String::with_capacity("mnemonic".len() + p.len());
     salt.push_str("mnemonic");
-    salt.push_str(p);
-    let salt = salt.nfkd().collect::<String>();
+    salt.push_str(p.0);
```
Alex6323: `let salt = ["mnemonic", p.0].into_iter().collect::<String>();`

src/keys/bip39.rs
semenov-vladyslav: Don't know. Historical reasons probably. Japanese wordlist uses U+3000 as space but it's still normalizes to U+0020. Separator can be a tab or a new-line char. Different separators result in different derived seeds. So it makes sense to have is as parameter.

src/keys/bip39.jp.rs
semenov-vladyslav: Initially Japanese wordlist used U+3000 as separator, but we need to normalize it anyway when deriving a seed. U+3000 normalizes to U+0020 which is a regular ASCII space that is used by default in English wordlist.

src/keys/bip39.rs
semenov-vladyslav: Removed

src/keys/bip39.en.rs
semenov-vladyslav: Not idea. Probably to not depend on an external crate. But we already depend on many other external crypto crates.

src/keys/bip39.rs
semenov-vladyslav: Done.

src/keys/bip39.rs
semenov-vladyslav: Agree about Zeroize. Do we need Eq? Why do you want to compare mnemonics?

src/keys/bip39.rs
semenov-vladyslav: Agree. Good point.

src/keys/bip39.rs
semenov-vladyslav: Right.

src/keys/bip39.rs
semenov-vladyslav: Right.

src/keys/bip39.rs
semenov-vladyslav: Done.

src/keys/bip39.rs
semenov-vladyslav: It's a bad idea to compare secrets. At least with the default Eq. It's not constant time and may lead some information about the secret. But why would you need Eq impl?

Alex6323: good counter question: it's probably quite handy for tests. Like you have maybe different ways of creating the mnemonic and you want to check whether that results in the same mnemonic.

src/keys/bip39.rs
Alex6323: wrote a bit more text down further down below ... short answer: probably good for tests? I agree with you they shouldn't be compared directly, but rather only the derived seed for authentication, right?

src/keys/bip39.rs
```diff
@@ -43,17 +165,43 @@ pub mod wordlist {
...
+        const fn new_unchecked(separator: char, words: &'a [&'a str; 2048]) -> Self {
+            Self { words, separator }
+        }
+
+        pub fn new(separator: char, words: &'a [&'a str; 2048]) -> Result<Self, Error> {
+            // normalize separator char
+            let mut s = String::new();
+            s.push(separator);
+            let s: String = s.nfkd().collect();
+            let mut s_chars = s.chars();
```
Alex6323: I think you can simplifiy this code (don't need to collect the iterator returned by `nfkd()`. Something like this should already work since `Decompositions<I>` implements an iterator over chars:
```
let mut s_chars = String::from(separator).nfkd();
```

src/keys/bip39.rs
```diff
@@ -5,24 +5,137 @@
...
+impl<'a> PartialEq<str> for MnemonicRef<'a> {
+    fn eq(&self, other: &str) -> bool {
+        self.0.eq(other)
+    }
+}
+
+#[derive(Clone, Zeroize, ZeroizeOnDrop)]
+pub struct Mnemonic(String);
+
+impl From<String> for Mnemonic {
```
Alex6323, line 78: could also be interesting to support wordlists in the form of `Vec<String>` and `[&'static str, 24]`.

semenov-vladyslav: We do not know separator at this point. Shall we use space by default?

src/keys/bip39.rs
semenov-vladyslav: Tests can use `as_ref()` to extract the inner value.

src/keys/bip39.rs
semenov-vladyslav: Mnemonic/Passphrase behave like String and impl AsRef<str>, and MnemonicRef/PassphraseRef behave like &str and impl Deref.

src/keys/bip39.rs
semenov-vladyslav: You do not normally compare seeds. For test purposes there's as_ref() to access inner value.

tests/bip39.rs
semenov-vladyslav: Done.

src/keys/bip39.rs
semenov-vladyslav: Done.

src/keys/bip39.rs
Alex6323: alright then :+1: ... can be closed!

src/keys/bip39.rs
Alex6323: okay :+1:

src/keys/bip39.rs
Alex6323: Naively I would just say yes, but maybe there's more to it then I can think of right now. Can you think of any problems with doing so?

semenov-vladyslav: Different separators will result in different seed. But this should be easily detected. I think we can use space with a warning in the comments.

src/keys/bip39.rs
semenov-vladyslav: Done.

src/keys/bip39.rs
semenov-vladyslav: Done.

src/keys/bip39.rs
semenov-vladyslav: Done.

src/keys/bip39.rs
semenov-vladyslav: Done, space is used as separator.

src/keys/bip39.rs
```diff
@@ -5,34 +5,216 @@
...
+    BadWordlistWord(String),
+    /// Word list contains duplicate words
+    UnsortedWordlist,
+    /// Separator is not in NFKD form
+    BadSeparator,
+}
+
+/// Reference to a normalized (unicode NFKD) mnemonic.
+#[derive(Clone, Copy, PartialEq, Eq)]
+pub struct MnemonicRef<'a>(&'a str);
```
Alex6323: Don't know why I didn't ask this in the other `Eq`/`PartialEq` discussion: but why do we make `MnemonicRef`s comparable, but not `Mnemonic`s? Didn't you argue it's not necessary to compare them? :thinking:

src/keys/bip39.rs
```diff
@@ -5,34 +5,216 @@
...
 
-    let mut salt = String::with_capacity("mnemonic".len() + p.len());
-    salt.push_str("mnemonic");
-    salt.push_str(p);
-    let salt = salt.nfkd().collect::<String>();
+/// Normalize the input string and use it as mnemonic.
+/// The resulting mnemonic should be verified against a given word list before deriving a seed from it.
+impl From<String> for Mnemonic {
+    fn from(mut unnormalized_mnemonic: String) -> Self {
+        let mnemonic = Self(unnormalized_mnemonic.chars().nfkd().collect());
```
Alex6323, line 80: I am actually astonished that this normalization does work for any `String` (i.e. it is infallible). What would a completely invalid string like 
"    abcdef   " (with leading and trailing whitespace) result in for example?

Alex6323: Why is not something necessary like this?:
```rust
impl TryFrom<String> for Mnemonic {
    type Error = Error;

    fn try_from(mut value: String) -> Result<Self, Self::Error> {
        // trim because empty spaces could create a different seed https://github.com/iotaledger/crypto.rs/issues/125
        let trimmed = value.trim();
        // first we check if the mnemonic is valid to give meaningful errors
        if let Err(err) = crypto::keys::bip39::wordlist::verify(trimmed, &crypto::keys::bip39::wordlist::ENGLISH) {
            value.zeroize();
            Err(crate::client::Error::InvalidMnemonic(format!("{err:?}")))
        } else {
            let mnemonic = trimmed.to_string();
            value.zeroize();
            Ok(Self(mnemonic))
        }
    }
}
```

Alex6323: oh ... I missed this line in the docs "Mnemonic should be verified against a given word list before deriving a seed from it." So we allow invalid `Mnemonic`s (but fail when we try to derive the actual seed from it), right?

src/keys/bip39.rs
semenov-vladyslav: Right, forgot to remove Eq derives.

src/keys/bip39.rs
semenov-vladyslav: This is unicode NFKD normalization. It is needed to have a unique representation (as bytestring) of the same string written in different unicode characters. This `from` conversion doesn't check for spaces and words. This mnemonic should be verified against the intended word list. I've added a doc comment regarding this.

src/keys/bip39.rs
semenov-vladyslav: Removed

src/keys/bip39.rs
```diff
@@ -5,34 +5,216 @@
...
+    BadWordlistWord(String),
+    /// Word list contains duplicate words
+    UnsortedWordlist,
+    /// Separator is not in NFKD form
+    BadSeparator,
+}
+
+/// Reference to a normalized (unicode NFKD) mnemonic.
+#[derive(Clone, Copy)]
+pub struct MnemonicRef<'a>(&'a str);
```
DaughterOfMars, line 40: Can you impl `ToOwned` here and `Borrow` for Mnemonic?

DaughterOfMars: Same with Password

src/keys/bip39.rs
Alex6323: Yeah, that's what I missed and saw eventually. So you think it's better to lazily verify the mnemonic (when actually needed for seed derivation) rather than eagerly in the `From` or then probably `TryFrom` impl?

semenov-vladyslav: Not really sure. You can derive seed from any string/mnemonic even if it's not a valid according to some standard word lists. It would probably be better to validate mnemonic, but at this point (in `from` and `try_from` we don't have any word list to validate against). I'm fine with letting users use mnemonics without validation if that's what they want.

src/keys/bip39.rs
semenov-vladyslav: I tried and failed. The major problem is with `ToOwned` which is already implemented for `T: Clone`. With `Borrow` there's generic lifetime argument to `MnemonicRef` which we need to know in advance. And `borrow` need to return reference to `MnemonicRef` not value.

src/keys/bip39.rs
Alex6323: okay, then I think in the SDK we have to wrap the `crypto.rs` type, which is fine, I guess and do the verification there. That's because the SDK doesn't allow users to use any string.

qrayven (approved):

src/keys/bip39.rs
Alex6323: can be closed if you want

DaughterOfMars (approved):

Alex6323 (approved):

src/keys/bip39.rs
semenov-vladyslav: Maybe, there should be `EnglishMnemonic` that uses the standard English word list. Or there can be `VerifiedMnemonic` that is returned by `wordlist::verify`.

PR #201, opened 2023-05-26
Typesafer slip10

semenov-vladyslav:
# Description of change

At the cost of a few additional traits we avoid runtime checks that can be caused by trying to derive using hardened/non-hardened segment for unsupporting key: ed25519::SecretKey can only derive hardened keys, secp256k1_ecdsa::PublicKey can only derive non-hardened keys. Also, unneeded Chain type is removed, it can be replaced by `Vec<u32>`, or `Vec<Hardened>`, or `Vec<NonHardened>`.

## Type of change

- [ ] Documentation Fix

A few tests became irrelevant, they were removed. The valid tests were adjusted to the new API.

## Change checklist

### DIFF
--- a/.changes/typesafer-slip10.md
+++ b/.changes/typesafer-slip10.md
@@ -1,0 +1,5 @@
+---
+"iota-crypto": minor
+---
+
+More static typesafety for SLIP10 implementation to avoid runtime checks.
--- a/.changes/typesafer-slip10.md
+++ b/.changes/typesafer-slip10.md
@@ -2,4 +2,4 @@
 "iota-crypto": minor
 ---
 
-More static typesafety for SLIP10 implementation to avoid runtime checks.
+More static type safety for SLIP10 implementation to avoid runtime checks.
--- a/src/keys/slip10.rs
+++ b/src/keys/slip10.rs
@@ -452,6 +452,7 @@ impl<K: hazmat::Derivable> TryFrom<&[u8; 65]> for Slip10<K> {
     }
 }
 
+/// Segment of a derivation chain.
 pub trait Segment: Copy + Into<u32> {
     fn is_hardened(self) -> bool;
     fn ser32(self) -> [u8; 4] {
@@ -461,9 +462,13 @@ pub trait Segment: Copy + Into<u32> {
     fn unharden(self) -> NonHardened;
 }
 
+/// Error indicating unexpected/invalid segment hardening.
+/// Some keys only accept certain segments: either hardened or non-hardened.
 #[derive(Clone, Copy, Debug, Eq, PartialEq)]
 pub enum SegmentHardeningError {
+    /// Input segment is hardened, expected non-hardened segment only.
     Hardened,
+    /// Input segment is non-hardened, expected hardened segment only.
     NonHardened,
 }
 
@@ -475,6 +480,7 @@ impl From<SegmentHardeningError> for crate::Error {
 
 const HARDEN_MASK: u32 = 1 << 31;
 
+/// `u32` type can represent both hardened and non-hardened segments.
 impl Segment for u32 {
     fn is_hardened(self) -> bool {
         self & HARDEN_MASK != 0
@@ -487,6 +493,7 @@ impl Segment for u32 {
     }
 }
 
+/// Type of hardened segments.
 #[derive(Clone, Copy, Debug, Eq, Hash, PartialEq, Serialize, Deserialize)]
 pub struct Hardened(u32);
 
@@ -519,6 +526,7 @@ impl Segment for Hardened {
     }
 }
 
+/// Type of non-hardened segments.
 #[derive(Clone, Copy, Debug, Eq, Hash, PartialEq, Serialize, Deserialize)]
 pub struct NonHardened(u32);
 
--- a/src/keys/slip10.rs
+++ b/src/keys/slip10.rs
@@ -292,8 +292,9 @@ impl<K: hazmat::IsSecretKey> From<&Seed> for Slip10<K> {
     }
 }
 
-impl<K: hazmat::IsPublicKey> Slip10<K>
+impl<K> Slip10<K>
 where
+    K: hazmat::IsPublicKey,
     K::SecretKey: hazmat::ToPublic,
 {
     pub fn from_extended_secret_key(esk: &Slip10<K::SecretKey>) -> Self {
@@ -304,8 +305,9 @@ where
     }
 }
 
-impl<K: hazmat::IsPublicKey> From<&Slip10<K::SecretKey>> for Slip10<K>
+impl<K> From<&Slip10<K::SecretKey>> for Slip10<K>
 where
+    K: hazmat::IsPublicKey,
     K::SecretKey: hazmat::ToPublic,
 {
     fn from(esk: &Slip10<K::SecretKey>) -> Self {
@@ -412,7 +414,7 @@ impl<K: hazmat::Derivable> Slip10<K> {
     {
         let mut data = [0u8; 33 + 4];
         data[..33].copy_from_slice(&self.calc_data(segment));
-        data[33..].copy_from_slice(&segment.ser32()); // ser32(i)
+        data[33..].copy_from_slice(&segment.ser32());
 
         let mut key = Self::new();
         HMAC_SHA512(&data, self.chain_code(), key.ext_mut());
@@ -445,8 +447,8 @@ pub trait Segment: Copy + Into<u32> {
 
 #[derive(Clone, Copy, Debug, Eq, PartialEq)]
 pub enum SegmentHardeningError {
-    SegmentHardened,
-    SegmentNonHardened,
+    Hardened,
+    NonHardened,
 }
 
 impl From<SegmentHardeningError> for crate::Error {
@@ -484,7 +486,7 @@ impl TryFrom<u32> for Hardened {
         if segment.is_hardened() {
             Ok(Hardened(segment))
         } else {
-            Err(SegmentHardeningError::SegmentNonHardened)
+            Err(SegmentHardeningError::NonHardened)
         }
     }
 }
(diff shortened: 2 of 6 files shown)


src/keys/slip10.rs
```diff
@@ -363,21 +390,28 @@ impl<K: hazmat::Derivable> Slip10<K> {
...
+        S: Segment,
+        K: hazmat::CalcData<S>,
+    {
+        K::calc_data(self.key_bytes(), segment)
+        // if hardened {
+        //     *self.key_bytes()
+        // } else {
+        //     debug_assert!(K::ALLOW_NON_HARDENED);
+        //     K::calc_non_hardened_data(self.key_bytes())
+        // }
```
Thoralf-M: Can be removed?

src/keys/slip10.rs
```diff
@@ -33,16 +30,20 @@ mod hazmat {
     pub trait IsPublicKey: Derivable {
         type SecretKey: IsSecretKey;
     }
+    pub trait ToPublic: IsSecretKey {
+        fn to_public(sk_bytes: &[u8; 33]) -> [u8; 33];
+    }
+    pub trait CalcData<S: Segment> {
+        fn calc_data(key_bytes: &[u8; 33], segment: S) -> [u8; 33];
```
Thoralf-M, line 52: What's `calc_data` doing?  Calculating the public key or what data?
The name isn't really helpful

src/keys/slip10.rs
```diff
@@ -399,88 +433,89 @@ impl<K: hazmat::Derivable> TryFrom<&[u8; 65]> for Slip10<K> {
...
-        let mut ss = self.0.clone();
-        ss.extend_from_slice(&o.as_ref().0);
-        Self(ss)
+impl TryFrom<u32> for Hardened {
+    type Error = ();
+    fn try_from(segment: u32) -> Result<Self, ()> {
+        if segment.is_hardened() {
+            Ok(Hardened(segment))
+        } else {
+            Err(())
```
Thoralf-M: Should return an actual Error and not just `()`?

src/keys/slip10.rs
```diff
@@ -399,88 +433,89 @@ impl<K: hazmat::Derivable> TryFrom<&[u8; 65]> for Slip10<K> {
...
-impl Default for Chain {
-    fn default() -> Self {
-        Chain::empty()
+impl TryFrom<u32> for NonHardened {
+    type Error = ();
+    fn try_from(segment: u32) -> Result<Self, ()> {
+        if !segment.is_hardened() {
+            Ok(NonHardened(segment))
+        } else {
+            Err(())
```
Thoralf-M: Same here, should return an actual Error

src/keys/slip10.rs
semenov-vladyslav: Sure, removed.

src/keys/slip10.rs
semenov-vladyslav: Agree. It calculates part of 'data' argument to be hashed in hmac which is either a secret key or a public key. It's implementation detail and it has to be `pub`.

src/keys/slip10.rs
semenov-vladyslav: Probably should use crate::Error for that matter.

semenov-vladyslav: What would be the best way to define an error?

semenov-vladyslav: Added `SegmentHardeningError`. Added `Slip10Error` pattern to `crypto::Error` enum.

kwek20 (approved):

DaughterOfMars (approved):

.changes/typesafer-slip10.md
```diff
@@ -0,0 +1,5 @@
+---
+"iota-crypto": minor
+---
+
+More static typesafety for SLIP10 implementation to avoid runtime checks.
```
Alex6323:
```suggestion
More static type safety for SLIP10 implementation to avoid runtime checks.
```
sorry ... I couldn't unsee it again.

src/keys/slip10.rs
```diff
@@ -399,88 +427,101 @@ impl<K: hazmat::Derivable> TryFrom<&[u8; 65]> for Slip10<K> {
...
+    fn unharden(self) -> NonHardened {
+        NonHardened(self & !HARDEN_MASK)
     }
+}
 
-    pub fn from_u32_hardened<I: IntoIterator<Item = u32>>(is: I) -> Self {
-        Self::from_u32(is.into_iter().map(|i| Segment::HARDEN_MASK | i))
-    }
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq, Serialize, Deserialize)]
+pub struct Hardened(pub u32);
```
Alex6323: hmm ... do we need the inner to be `pub` when we have a fallible `TryFrom`? This could be misused and create an invalid `Hardened`, right?

DaughterOfMars:
Also, clippy and tests failing

semenov-vladyslav:
Clippy fails in ternary_buf tests due to redundant `a.clone()`. I've added `#[allow(clippy::redundant_clone)]`, it seems to have worked locally but not on CI. Need to investigate further.

Sorry, I mixed up branches, it's fixed in feat/bip39-safety-wrappers branch.

.changes/typesafer-slip10.md
semenov-vladyslav: Fixed

src/keys/slip10.rs
semenov-vladyslav: Right. Copy-paste error, didn't notice myself. Fixed, inner values are private.

DaughterOfMars: Good point. This should probably be `pub(crate)`

src/keys/slip10.rs
```diff
@@ -253,18 +285,22 @@ impl<K: hazmat::IsSecretKey> From<&Seed> for Slip10<K> {
     }
 }
 
-impl<K: hazmat::IsPublicKey> Slip10<K> {
+impl<K: hazmat::IsPublicKey> Slip10<K>
+where
+    K::SecretKey: hazmat::ToPublic,
+{
```
Alex6323: I'ld probably put `K: hasmat::IsPublicKey` under the `where` clause as well, which makes the bounds easier to read IMO.

src/keys/slip10.rs
```diff
@@ -253,18 +285,22 @@ impl<K: hazmat::IsSecretKey> From<&Seed> for Slip10<K> {
...
         k.ext[33..].copy_from_slice(esk.chain_code());
         k
     }
 }
 
-impl<K: hazmat::IsPublicKey> From<&Slip10<K::SecretKey>> for Slip10<K> {
+impl<K: hazmat::IsPublicKey> From<&Slip10<K::SecretKey>> for Slip10<K>
+where
+    K::SecretKey: hazmat::ToPublic,
+{
```
Alex6323: here as well

src/keys/slip10.rs
```diff
@@ -363,21 +390,22 @@ impl<K: hazmat::Derivable> Slip10<K> {
...
+    fn derive_child_key<S>(&self, segment: S) -> Self
+    where
+        S: Segment,
+        K: hazmat::CalcData<S>,
+    {
         let mut data = [0u8; 33 + 4];
-        data[..33].copy_from_slice(&self.calc_data(segment.is_hardened()));
-        data[33..].copy_from_slice(&segment.bs()); // ser32(i)
+        data[..33].copy_from_slice(&self.calc_data(segment));
+        data[33..].copy_from_slice(&segment.ser32()); // ser32(i)
```
Alex6323: is this a comment?

src/keys/slip10.rs
```diff
@@ -399,88 +427,101 @@ impl<K: hazmat::Derivable> TryFrom<&[u8; 65]> for Slip10<K> {
...
 
-impl From<u32> for Segment {
-    fn from(i: u32) -> Self {
-        Self(i)
-    }
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub enum SegmentHardeningError {
+    SegmentHardened,
+    SegmentNonHardened,
 }
```
Alex6323: probably can remove the `Segment...` prefix on the ~variants~ type. Ignore and close if you disagree. I don't care so much.

src/keys/slip10.rs
semenov-vladyslav: Agree. Done.

src/keys/slip10.rs
semenov-vladyslav: It is. Long time ago it meant something. Removed.

src/keys/slip10.rs
semenov-vladyslav: Agree. Removed.

src/keys/slip10.rs
semenov-vladyslav: Sure. Fixed

DaughterOfMars (approved):

semenov-vladyslav:
Shall we export `hazmat` traits? Currently they are public (to make rust happy) but not exported thus they cannot be imported and used in generic contexts. It might be useful to have something like this in an external crate:

```rust
fn f<K: slip10::Derivable>(ext: slip10::Slip10<K>);
fn g<K: slip10::IsSecretKey>(ext: slip10::Slip10<K>);
```

Currently it's not possible, so the external crate has to workaround:

```rust
fn fEd25519(ext: slip10::Slip10<Ed25519::SecretKey>);
fn fSecp256k1Secret(ext: slip10::Slip10<Secp256k1::SecretKey>);
fn fSecp256k1Public(ext: slip10::Slip10<Secp256k1::PublicKey>);
```

From the other side making `hazmat` traits public would allow users to define their own unsafe implementations.

The current approach: do not export traits, if needed they can be exported in a patch update.

DaughterOfMars:
If you want to allow users access to the traits without allowing them to make unsafe impls, then use a [sealed pattern](https://rust-lang.github.io/api-guidelines/future-proofing.html).

semenov-vladyslav:
Thanks. Added Sealed trait, exported all other hazmat traits, added test.

DaughterOfMars (approved):

Alex6323 (approved):

PR #205, opened 2023-06-14
Derive/impl more traits for Signatures

DaughterOfMars:
# Description of change

`iota_sdk` could use these types more effectively if they have more impls.

## Type of change

- Enhancement (a non-breaking change which adds functionality)


src/keys/slip10.rs
```diff
@@ -122,7 +122,7 @@ pub mod secp256k1 {
                 if scalar.is_zero().into() {
                     false
                 } else {
-                    key_bytes[1..].copy_from_slice(&scalar.to_bytes());
+                    key_bytes[1..].copy_from_slice(&*scalar.to_bytes());
```
thibault-martinez: This doesn't look clippy friendly

src/signatures/ed25519.rs
```diff
@@ -62,7 +62,15 @@ impl SecretKey {
     }
 }
 
+impl Eq for SecretKey {}
+impl PartialEq for SecretKey {
+    fn eq(&self, other: &Self) -> bool {
+        self.0.as_ref() == other.0.as_ref()
```
semenov-vladyslav: It's not safe to compare secrets using the usual equality. The safe way should be [constant time](https://docs.rs/subtle/latest/subtle/trait.ConstantTimeEq.html).

thibault-martinez (changes requested):
Need a .change

src/signatures/ed25519.rs
```diff
@@ -148,3 +158,21 @@ impl Signature {
         Self(ed25519_zebra::Signature::from(bs))
     }
 }
+
+impl PartialOrd for Signature {
```
semenov-vladyslav, line 155: Why do we need to compare signatures?

src/signatures/secp256k1_ecdsa.rs
```diff
@@ -1,12 +1,13 @@
...
 // SPDX-License-Identifier: Apache-2.0
 
+use core::cmp::Ordering;
 use core::convert::TryFrom;
 use core::hash::{Hash, Hasher};
 
 use zeroize::{ZeroizeOnDrop, Zeroizing};
 
-#[derive(ZeroizeOnDrop)]
+#[derive(Clone, ZeroizeOnDrop, Eq, PartialEq)]
```
semenov-vladyslav: Eq and PartialEq are not constant time.

semenov-vladyslav (commented):
Maybe we should use subtle crate if we need to compare secret keys. It feels like misuse to compare keys.

src/signatures/ed25519.rs
DaughterOfMars: This implies that `iota_sdk` is misusing these values, because we have to compare/hash them for various usages.

semenov-vladyslav: What are these use cases? I thought secret keys were stored in stronghold

DaughterOfMars: They are part of a UTXO.

semenov-vladyslav: Definitely not secret keys.

DaughterOfMars: No, just Public keys and Signatures. I think I misread your comments. Are the changes to those two types acceptable? I will revert the changes to SecretKeys.

src/keys/slip10.rs
DaughterOfMars: Weird, my workspace doesn't like it as-is. Oh well.

src/signatures/ed25519.rs
DaughterOfMars: Addressed above. We use it as part of UTXO.

src/signatures/secp256k1_ecdsa.rs
DaughterOfMars: Reverted

src/signatures/ed25519.rs
semenov-vladyslav: Yes, right

src/signatures/ed25519.rs
semenov-vladyslav: Alright. I thought that signature and public key **bytes** (not types) are compared/ordered as part of UTxO. Corresponding types should probably respect order in binary representation.

DaughterOfMars: Yes, we use the bytes. I had hoped that we could use the representation from crypto to avoid converting back-and-forth, but if you think that's unwise then I will abort.

thibault-martinez (changes requested):
.change

src/signatures/ed25519.rs
thibault-martinez: Has the comment been addressed?

src/signatures/ed25519.rs
thibault-martinez: Has the comment been addressed?

semenov-vladyslav: Yes

DaughterOfMars: You sure? There's still the question about the advisability of this change.

semenov-vladyslav: The current implementation via byte array is correct. It's possible to access inner signature fields r and s to avoid copying. Also, order on Signature type is total. I'd do it like this:

```rust
impl PartialOrd for Signature {
    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
        Some(self.cmp(other))
    }
}

impl Ord for Signature {
    fn cmp(&self, other: &Self) -> Ordering {
        let r_cmp = self.0.r_bytes().cmp(other.0.r_bytes());
        let s_cmp = self.0.s_bytes().cmp(other.0.s_bytes());
        r_cmp.then(s_cmp)
    }
}
```

To enable access to r and s bytes we need to update ed25519-zebra to version "4.0".

semenov-vladyslav: It would be nice to ensure/test property `for all s1, s2: [u8; 64] => s1.cmp(s2) == Signature::from_bytes(s1).cmp(&Signature.from_bytes(s2))`. This should hold as ed25519_zebra-v4.0 uses [ed25519::Signature](https://github.com/RustCrypto/signatures/blob/master/ed25519/src/lib.rs#L303) which keeps r and s components as bytes without and normalization.

src/signatures/secp256k1_ecdsa.rs
```diff
@@ -138,7 +140,12 @@ impl Hash for PublicKey {
     }
 }
 
-pub struct Signature(k256::ecdsa::Signature, k256::ecdsa::RecoveryId);
+#[derive(Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub struct Signature(
+    k256::ecdsa::Signature,
+    #[serde(with = "serde_recovery_id")] k256::ecdsa::RecoveryId,
```
semenov-vladyslav: `cfg_attr(feature = "serde", ...)`

src/signatures/secp256k1_ecdsa.rs
```diff
@@ -190,6 +197,48 @@ impl Signature {
...
+        D: serde::Deserializer<'de>,
+    {
+        use k256::ecdsa::RecoveryId;
+        use serde::Deserialize;
+        RecoveryId::from_byte(u8::deserialize(d)?)
+            .ok_or_else(|| serde::de::Error::custom(format!("invalid recovery byte (max {})", RecoveryId::MAX)))
+    }
+}
+
+impl PartialOrd for Signature {
```
semenov-vladyslav, line 220: Order is total:

```rust
impl PartialOrd for Signature {
    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
        Some(self.cmp(other))
    }
}

impl Ord for Signature {
    fn cmp(&self, other: &Self) -> Ordering {
        let (r1, s1) = self.0.split_bytes();
        let (r2, s2) = other.0.split_bytes();
        r1.cmp(&r2).then(s1.cmp(&s2)).then(self.1.cmp(&other.1))
    }
}
```

semenov-vladyslav (changes requested):

src/signatures/ed25519.rs
semenov-vladyslav: Yes, we do not compare/hash secret keys.

src/signatures/ed25519.rs
semenov-vladyslav: Signature cmp uses internal r/s values to avoid copying. Signature cmp respects corresponding byte ordering. Added tests.

src/signatures/secp256k1_ecdsa.rs
semenov-vladyslav: Done

src/signatures/secp256k1_ecdsa.rs
semenov-vladyslav: Done.

semenov-vladyslav (approved):

thibault-martinez (approved):

PR #211, opened 2023-07-06
BIP44

semenov-vladyslav:
# Description of change

Add support for [BIP-44](https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki) chains for slip10 module.

## Type of change

- [ ] Documentation Fix

## Change checklist


src/keys/slip10.rs
```diff
@@ -172,6 +192,19 @@ pub mod secp256k1 {
         }
     }
 
+    impl Bip44IntoIter for secp256k1_ecdsa::SecretKey {
```
DaughterOfMars: Feels like this should be implemented on `Bip44` with a generic type, and use `&self`.

semenov-vladyslav: Sure. It would be nice to have something like:

```rust
trait ToChain {
    type Chain<K>;
    fn to_chain<K>(&self) -> Self::Chain<K>;
}
```

But the impl is impossible without generic specializations:

```rust
impl ToChain for Bip44 {
    type Chain<ed25519::SecretKey> = [Hardened; 5];
    type Chain<secp256k1_ecdsa::SecretKey> = [u32; 5];
    fn to_chain<ed25519::SecretKey>(&self) -> Chain<ed25519::SecretKey>;
    fn to_chain<secp256k1_ecdsa::SecretKey>(&self) -> Chain<secp256k1_ecdsa::SecretKey>;
}
```

The best I can do is as follows:

```rust
pub trait ToChain<C>: Sealed {
    type Chain;
    fn to_chain(pre_chain: &C) -> Self::Chain;
}
```

and impl it for `ed25519::SecretKey` and `secp256k1_ecdsa::SecretKey`.

semenov-vladyslav: Maybe you have better ideas.

DaughterOfMars: Why not just
```rust
pub trait ToChain<K>: Sealed {
    type Chain;
    fn to_chain(&self) -> Self::Chain;
}

impl ToChain<ed25519::SecretKey> for Bip44 {
    type Chain = [Hardened; 5];
    ...
}

impl ToChain<secp256k1_ecdsa::SecretKey> for Bip44 {
    type Chain = [u32; 5];
    ...
}
```

semenov-vladyslav: It would be difficult to call `bip44.to_chain()` for a specific key type. Need to resolve for which key: `<Bip44 as ToChain<ed25519::SecretKey>>::to_chain(bip44)`.

DaughterOfMars: Yes, or it can be resolved via bounds. Another option is to have the traits implemented on the keys and have a fn defined for `Bip44` like so:

```rust
fn to_chain<K: ToChain>(&self) -> K::Chain {...}
```

semenov-vladyslav: Updated. Added `Bip44::to_chain`.

src/keys/bip39.rs
```diff
@@ -247,12 +277,24 @@ impl fmt::Debug for Passphrase {
...
+    }
+}
+
 impl AsRef<[u8; 64]> for Seed {
     fn as_ref(&self) -> &[u8; 64] {
         &self.0
     }
 }
 
+impl AsRef<[u8]> for Seed {
```
DaughterOfMars, line 286: I would probably stick to the above impl. Having both will make this unnecessarily difficult to use.

src/keys/bip39.rs
```diff
@@ -247,12 +277,24 @@ impl fmt::Debug for Passphrase {
 #[derive(Clone, Zeroize, ZeroizeOnDrop)]
 pub struct Seed([u8; 64]);
 
+impl Seed {
+    pub fn as_bytes(&self) -> &[u8; 64] {
```
DaughterOfMars: Maybe call this `inner`?

src/keys/slip10.rs
```diff
@@ -632,15 +628,23 @@ pub struct Bip44 {
 }
 
 impl Bip44 {
+    pub fn builder(self) -> Bip44Builder {
+        Bip44Builder(self)
```
DaughterOfMars: This is a very odd pattern. Usually a builder is used when you cannot default the base type.

src/keys/slip10.rs
```diff
@@ -82,16 +88,29 @@ pub mod ed25519 {
...
     }
 
-    impl CalcData<Hardened> for ed25519::SecretKey {
+    impl WithSegment<Hardened> for ed25519::SecretKey {
         fn calc_data(key_bytes: &[u8; 33], _segment: Hardened) -> [u8; 33] {
             *key_bytes
         }
     }
+
+    impl ToChain<Bip44> for ed25519::SecretKey {
```
DaughterOfMars: Given that this is not extensible, why have the generic type?

src/keys/bip39.rs
semenov-vladyslav: Right. I thought to make seed generic in `Slip10::from_seed<S: AsRef<[u8]>>(seed: &S)`. I'll remove `AsRef<[u8; 64]>` impl then - there's `fn to_bytes(&self) -> &[u8; 64]` as a replacement.

src/keys/bip39.rs
semenov-vladyslav: For keys we use `fn to_bytes(&self) -> [u8; LEN]` or `fn as_slice(&self) -> &[u8]`. Wanted to make something similar. The naming is a bit inconsistent. Not sure this function is needed at all.

src/keys/slip10.rs
semenov-vladyslav: `Bip44` doesn't `impl Default`, the idea is to be able to easily convert an instance into a builder and modify/update it via the builder.

semenov-vladyslav: `Bip44` doesn't `impl Default`, the idea is to be able to easily convert an instance into a builder and modify/update it via the builder.

src/keys/slip10.rs
semenov-vladyslav: To support other BIP32 chains. BIP44 chain has a fixed format, other BIP32 chains can be shorter/longer. Unlikely that we will need other chains.

src/keys/bip39.rs
DaughterOfMars: I would definitely keep the function. I think that for transparent wrappers like this `inner` makes the most sense. My second pick would be just `bytes`. `to_bytes` implies that it consumes the value, but this is by ref.

src/keys/slip10.rs
DaughterOfMars: Yes, but you take a value of `Bip44` in order to create the builder, which begs the question of how you create that without the builder. Catch-22.

src/keys/slip10.rs
DaughterOfMars: Sure, but this is a local, non-extensible trait so it's pointless.

src/keys/slip10.rs
```diff
@@ -588,3 +664,162 @@ impl Segment for NonHardened {
...
+pub struct Bip44Builder(Bip44);
+
+impl Bip44Builder {
+    pub fn new() -> Self {
+        Self(Bip44::from([0, 0, 0, 0, 0]))
+    }
+    pub fn purpose(mut self, s: u32) -> Self {
+        self.0.purpose = s;
+        self
+    }
```
Thoralf-M: Shouldn't the purpose be hardcoded to 44 (or actually it's hardened value?)?
https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki#purpose

src/keys/slip10.rs
semenov-vladyslav: Bip44 value can be created from Bip44Builder or from [u32; 5].

src/keys/slip10.rs
semenov-vladyslav: I think the question here should be whether we want to support other Bip32-compatible chains or not. If so, then ToChain shouldn't be Sealed probably. If not, then indeed there's no point in generics, but the trait name should reflect that, eg. Bip44ToChain or so.

src/keys/slip10.rs
semenov-vladyslav: Should Bip44::purpose also be private? Should impl From<[u32; 5]> for Bip44 assert that purpose == 44? Or should it be TryFrom?

src/keys/slip10.rs
DaughterOfMars: Sure, but if you already have a Bip44 then why do you need the builder?

src/keys/slip10.rs
DaughterOfMars: Correct. If you want to allow future impls, then unseal it.

src/keys/bip39.rs
semenov-vladyslav: Renamed to `bytes`.

src/keys/slip10.rs
semenov-vladyslav: For example, [SLIP-13](https://github.com/satoshilabs/slips/blob/master/slip-0013.md), [SLIP-16](https://github.com/satoshilabs/slips/blob/master/slip-0016.md), [SLIP-17](https://github.com/satoshilabs/slips/blob/master/slip-0017.md) use BIP-32 compatible chains.

src/keys/slip10.rs
Thoralf-M: If it can fail it should be try_from
Or another idea, maybe just rename it to Bip32? Then it's not limited to 44 (but with the names of the fields it seems to be somehow a mix or at least with Bip43)

src/keys/slip10.rs
semenov-vladyslav: Maybe, Bip44 deserves its own module and feature?

DaughterOfMars: Probably, because it's destined to be a shared type.

src/keys/slip10.rs
semenov-vladyslav: To construct a different bip44 chain. It can be done with clone and modifying a corresponding field, or it can be done with builder.

src/keys/slip10.rs
Thoralf-M: > Should Bip44::purpose also be private?

If we always only want to use it for Bip44 I would say yes, no point in having the option to set something else then

src/keys/slip10.rs
DaughterOfMars: That seems overly complex, why not just have accessor methods?

src/keys/slip10.rs
semenov-vladyslav: Removed it. Now it's fixed constant.

src/keys/slip10.rs
semenov-vladyslav: Removed.

src/keys/bip44.rs
```diff
@@ -0,0 +1,219 @@
...
+            bip44_chain.change,
+            bip44_chain.address_index,
+        ]
+    }
+}
+
+pub struct Bip44Builder(Bip44);
+
+impl Bip44Builder {
+    pub fn new() -> Self {
```
DaughterOfMars: I would probably replace this with `#[derive(Default)]`

DaughterOfMars: Or a custom impl of `Default`, if you don't want to derive it for `Bip44`. Though now I am wondering about the purpose of the builder, since it has no validation method. You could just move these setters to the `Bip44` type.

semenov-vladyslav: Removed Builder. Moved setters into Bip44. Derive Default.

src/keys/bip44.rs
```diff
@@ -60,6 +60,30 @@ pub struct Bip44 {
 impl Bip44 {
     pub const PURPOSE: u32 = 44;
 
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    pub fn coin_type(mut self, s: u32) -> Self {
```
DaughterOfMars: Now that these aren't part of a builder, they should be named `with_x`.

semenov-vladyslav: Renamed.

src/keys/slip10.rs
semenov-vladyslav: Moved bip44 into its own module. Added "bip44" feature.

DaughterOfMars (approved):

thibault-martinez (approved):

PR #221, opened 2024-01-16
Add ed25519 public key bytes wrapper

DaughterOfMars:
# Description of change

This PR adds a public key bytes wrapper that does not require validation to exist.

## Type of change

- [ ] Documentation Fix

Modified a few existing tests to use the new type.

## Change checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "iota-crypto"
-version = "0.23.1"
+version = "0.23.0"
 license = "Apache-2.0"
 authors = [ "IOTA Stiftung" ]
 edition = "2021"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "iota-crypto"
-version = "0.23.0"
+version = "0.23.1"
 license = "Apache-2.0"
 authors = [ "IOTA Stiftung" ]
 edition = "2021"
--- a/src/hashes/ternary/kerl/bigint/i384/mod.rs
+++ b/src/hashes/ternary/kerl/bigint/i384/mod.rs
@@ -168,6 +168,12 @@ impl PartialEq for I384<BigEndian, U8Repr> {
 
 impl PartialOrd for I384<BigEndian, U8Repr> {
     fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+impl Ord for I384<BigEndian, U8Repr> {
+    fn cmp(&self, other: &Self) -> Ordering {
         use Ordering::*;
 
         let mut zipped_iter = self.inner.iter().zip(other.inner.iter());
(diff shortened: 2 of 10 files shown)


semenov-vladyslav:
What is the purpose of such wrapper? In which way the wrapper "does not require validation"?

In my opinion, it adds extra code duplication without clear benefits. Could you please provide a non-trivial use-case example demonstrating benefits of the wrapper?

DaughterOfMars:
It does not require validation of bytes for the type to be constructed, unlike `PublicKey`. This allows us to unpack a public key without guaranteeing anything other than the length of the bytes. We need this in shimmer because currently invalid bytes are allowed syntactically (something we hope to change for 2.0).

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "iota-crypto"
-version = "0.23.0"
+version = "0.23.1"
```
semenov-vladyslav: We do not update version manually. See [.changes](https://github.com/iotaledger/crypto.rs/blob/dev/.changes/readme.md).

src/hashes/ternary/kerl/bigint/i384/mod.rs
```diff
@@ -168,6 +168,12 @@ impl PartialEq for I384<BigEndian, U8Repr> {
 
 impl PartialOrd for I384<BigEndian, U8Repr> {
     fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        Some(self.cmp(other))
```
semenov-vladyslav, line 171: Maybe such changes belong to a different PR?

src/signatures/ed25519.rs
```diff
@@ -136,6 +136,91 @@ impl Hash for PublicKey {
...
+
+impl PublicKeyBytes {
+    pub const LENGTH: usize = PublicKey::LENGTH;
+
+    pub fn verify(&self, sig: &Signature, msg: &[u8]) -> crate::Result<bool> {
+        Ok(self.into_public_key()?.verify(sig, msg))
+    }
+
+    pub fn into_public_key(self) -> crate::Result<PublicKey> {
+        Ok(PublicKey(self.0.try_into().map_err(|_| {
```
semenov-vladyslav: `self.0.try_into().map_err(...).map(PublicKey)`?

tests/ed25519.rs
```diff
@@ -24,13 +24,13 @@ fn test_zip215() -> crypto::Result<()> {
     for tv in tvs.iter() {
         let mut pkb = [0; PUBLIC_KEY_LENGTH];
         hex::decode_to_slice(tv.public_key, &mut pkb as &mut [u8]).unwrap();
-        let pk = PublicKey::try_from_bytes(pkb)?;
+        let pk = PublicKeyBytes::from_bytes(pkb);
```
semenov-vladyslav: These changes remove tests for `PublicKey`. Maybe, duplicate it like this:

```rust
let pk = PublicKey::try_from_bytes(pkb)?;
let pkb = PublicKeyBytes::from_bytes(pkb);
...
assert!(pk.verify(&sig, ms));
assert!(pkb.verify(&sig, ms));
```

semenov-vladyslav (changes requested):

Cargo.toml
DaughterOfMars: right, thanks

src/hashes/ternary/kerl/bigint/i384/mod.rs
DaughterOfMars: just figured I'd fix the clippy issues while I was there

tests/ed25519.rs
DaughterOfMars: hm, ok

semenov-vladyslav (approved):
Looks good.

thibault-martinez (approved):

